PR #8, opened 2022-08-25
feat(substreams-helper,erc20-price): price lib mvp

### LINKED ISSUE #7: substreams-helper: add price.rs
reference: https://github.com/0xbe1/price-rs

0xbe1:
I decide that price lib (price.rs) should depend on substreams, and it can only be used for substreams. Then we have a price module (erc20-price) that retrieves and caches all erc20 prices.

why? because substreams cannot run arbitrary rust lib. especially for contract calls, only eth_call in substreams-ethereum is supported

close #7


erc20-price/substreams.yaml
```diff
@@ -0,0 +1,27 @@
+specVersion: v0.1.0
```
this-username-is-taken, line 1: I thought you were going to make this a generic (non-substream) Rust crate?

substreams-helper/src/math.rs
```diff
@@ -26,17 +26,12 @@ pub fn decimal_from_hex_be_bytes(price_bytes: &Vec<u8>) -> BigDecimal {
         .with_prec(100);
 }
 
-pub fn exponent_to_big_decimal(decimals: &BigInt) -> BigDecimal {
+pub fn exponent_to_big_decimal(decimals: u8) -> BigDecimal {
```
this-username-is-taken, line 29: Why modify the old one? We probably want both

substreams-helper/src/price.rs
```diff
@@ -0,0 +1,52 @@
+use std::ops::Div;
+use std::str::FromStr;
+
+use bigdecimal::BigDecimal;
+use serde::Deserialize;
+use substreams::Hex;
+
+use crate::{math, rpc};
+
+pub enum Network {
```
this-username-is-taken, line 10: Should probably put shared enums in a separate module

substreams-helper/src/rpc.rs
```diff
@@ -0,0 +1,64 @@
...
+    let mut data = Vec::with_capacity(8 + args.len() * 32);
+    data.extend(method_sig);
+    for arg in args {
+        data.extend(vec![0u8].repeat(32 - arg.len()));
+        data.extend(arg);
+    }
+    return data;
+}
+
+// "name()" -> "06fdde03"
```
this-username-is-taken, line 27: Should change this to a documentation test: https://doc.rust-lang.org/rustdoc/write-documentation/documentation-tests.html#documentation-tests

substreams-helper/src/price.rs
```diff
@@ -0,0 +1,52 @@
...
+#[derive(Deserialize)]
+struct NetworkConfig {
+    yearn_lens_oracle: String,
+    usdc_decimals: u8,
+}
+
+pub fn get_price(network: Network, token_address: &str) -> Result<BigDecimal, String> {
+    let config = Config {
+        ethereum: NetworkConfig {
+            yearn_lens_oracle: "83d95e0d5f402511db06817aff3f9ea88224b030".to_string(),
```
this-username-is-taken, line 28: Should make all addresses into named constants and put them in a module

erc20-price/src/lib.rs
```diff
@@ -0,0 +1,17 @@
...
+
+use pb::erc20_price::Erc20Price;
+use substreams_helper::price;
+use substreams_helper::price::Network;
+
+mod pb;
+
+#[substreams::handlers::map]
+fn eth_price(block: eth::Block) -> Result<Erc20Price, substreams::errors::Error> {
+    let eth_price = price::get_price(Network::Ethereum, "c02aaa39b223fe8d0a0e5c4f27ead9083c756cc2").unwrap();
```
this-username-is-taken, line 12: same here

erc20-price/substreams.yaml
0xbe1: Not possible though, since all contract calls need to go through substreams-ethereum eth_call, otherwise it cannot run in substreams

substreams-helper/src/math.rs
0xbe1: Bc we are not calculating 10^bigint. The function is only for 10^token_decimals afaik

substreams-helper/src/price.rs
0xbe1: By a shared module do you mean something like types.rs? Not doing that because it is not shared yet, will move to a shared module when it has to

substreams-helper/src/rpc.rs
```diff
@@ -0,0 +1,64 @@
+use substreams::Hex;
+use substreams_ethereum::{pb::eth as ethpb, rpc};
+use tiny_keccak::{Hasher, Keccak};
+
+#[derive(Debug)]
+pub struct RpcCallParams {
+    pub to: Vec<u8>,
```
this-username-is-taken, line 7: Any particular reason that `to` is a byte array while `method` is a String? I think we should keep them consistent

substreams-helper/src/price.rs
0xbe1: How about putting all config like this in a toml file?

substreams-helper/src/price.rs
```diff
@@ -0,0 +1,52 @@
...
+pub enum Network {
+    Ethereum,
+}
+
+#[derive(Deserialize)]
+struct Config {
+    ethereum: NetworkConfig,
+}
+
+#[derive(Deserialize)]
```
this-username-is-taken, line 19: why does this need to be serializable?

substreams-helper/src/price.rs
```diff
@@ -0,0 +1,52 @@
...
+struct NetworkConfig {
```
this-username-is-taken, line 20: Can you add a comment on what a `NetworkConfig` encapsulates?

substreams-helper/src/rpc.rs
0xbe1: Because bytes are for machine while strings are for human. From a user perspective, it is easiest to just put a string for method name; while for contract address, user never have to manually input, that field, therefore I use bytes, which is more native to substreams

substreams-helper/src/price.rs
0xbe1: Gonna remove. It was there simply bc I used to read config from toml file. That way serializable is required

substreams-helper/src/math.rs
this-username-is-taken: I understand we are not using it here, but we might need to use it elsewhere. (This code snippet is copied from the Uniswap v3 substream and they use `BigInt` pretty extensively throughout their codebase)

substreams-helper/src/price.rs
this-username-is-taken: well this is an enum that we know for sure will be shared

substreams-helper/src/rpc.rs
0xbe1: should be in a separate PR to set up rustdoc ci

substreams-helper/src/rpc.rs
this-username-is-taken: should we change const addresses to byte arrays as well then? We should decide how we want to treat addresses and have it be consistent across the code base

substreams-helper/Cargo.toml
```diff
@@ -17,7 +17,12 @@ prost = "^0.11.0"
 num-bigint = "0.4"
 bigdecimal = "0.3"
 pad = "0.1"
+substreams = "~0.0.20"
 substreams-ethereum = "~0.2.0"
+serde_json = "1.0"
```
this-username-is-taken, line 22: I guess we can remove this as well now?

substreams-helper/src/math.rs
0xbe1: most of the time, `decimals` is from `decimals: u64` of `Erc20Token`.

- first of all, I feel it'd be better to just do `decimals: u8` - that's already 0-255, large enough for decimals
- with `exponent_to_big_decimal(decimals: u8)` we don't need to do the unit -> bigint conversion right?

but anyway if you want to keep both, I can keep both, but will that make someone misuse them?

substreams-helper/src/rpc.rs
0xbe1: good point, i end up using this:

```
const CONFIG: Config = Config {
    ethereum: NetworkConfig {
        yearn_lens_oracle: hex!("83d95e0d5f402511db06817aff3f9ea88224b030"),
        usdc_decimals: 6,
    },
};
```

`hex!("83d95e0d5f402511db06817aff3f9ea88224b030")` produces a byte array

substreams-helper/src/rpc.rs
this-username-is-taken: sounds good

substreams-helper/src/math.rs
this-username-is-taken: yeah, that sounds good. Let's keep decimals u8 then

this-username-is-taken:
Great work!

PR #9, opened 2022-08-26
fix(erc721): fix the build and bump dependencies

0xbe1:
close https://github.com/messari/substreams/issues/10


this-username-is-taken (approved):
Thanks!

PR #39, opened 2022-08-31
feat(substreams-helper): add chainlink to price lib

0xbe1:


substreams-helper/src/price.rs
```diff
@@ -21,16 +23,18 @@ struct Config {
 /// Price lib config for each supported network
 struct NetworkConfig {
     yearn_lens_oracle: [u8; 20],
+    chainlink_feed_registry: [u8; 20],
```
this-username-is-taken, line 26: can move `Address` into `substreams-helper` and change `[u8; 20]` to `Address`

this-username-is-taken (approved):
LGTM 🚀

substreams-helper/src/price.rs
0xbe1: i am thinking both [u8; 20] and Vec<u8> should be considered `Address`, not sure what's the best way, but let me delay this a bit

PR #46, opened 2022-09-03
feat(erc20-price): uncheck src/pb and src/abi and let build generate

0xbe1:
Currently build fails because we need to run `substreams codegen` in CI. Still figuring out how to install substreams cli and buf cli correctly in GitHub Actions.

Though not a blocker for merge bc we can run `make build-all` locally to verify

---

If this looks good, gonna port this to other substreams :)


this-username-is-taken:
Looks great!

Issue #25, opened 2022-08-31
lib should not version control Cargo.lock


0xbe1:
https://doc.rust-lang.org/cargo/faq.html#why-do-binaries-have-cargolock-in-version-control-but-not-libraries

PR #51, opened 2022-09-10
chore(substreams-helper): remove cargo.lock from lib vcs

0xbe1:
close https://github.com/messari/substreams/issues/25


0xbe1:
@Ryanmtate Want to get your input here - is this considered good practice in Rust?

Ryanmtate:
Since the crates will likely be libraries within the workspace, I think we could ignore the `Cargo.lock` file. Now that we're adding a workspace, we can ignore the root lock file for the workspace.

If we decide to add a binary application later on, we may wish to commit the lock file.

See: https://github.com/rust-lang/cargo/issues/6172

Ryanmtate (approved):

Ryanmtate (approved):

Issue #20, opened 2022-08-31
move src/pb and src/abi out of vcs


0xbe1:
- [x] #45
- [x] #47
- [x] #48
- [x] #49

PR #60, opened 2022-09-15
codegen: standardize abi and proto codegen build scripts

Ryanmtate:
Places generated abi and pb rust files into the local crate `target/` directory, thereby ignoring the generated code within the repo. Codegen works by looking up the generated file names, which should be fully qualified and include a major version number, e.g. `example.type.v1`. The substreams codegen places the pb files into a `src/pb/` folder using the file name: `example.type.v1.rs`. The codegen will move these generated files into the target folder and replace the file name as a versioned module. e.g.

```rust
// Filename: `src/pb/example_type.rs`

#[rustfmt::skip]
#[path = "../../target/pb/example.type.v1.rs"]
pub mod v1;

#[rustfmt::skip]
#[path = "../../target/pb/example.type.v2.rs"]
pub mod v2;
```

A root `src/pb.rs` module file is also created that contains the top level type:

```rust
pub mod example_type;
```

Notice the filename and module names are underscored. This is handled by the codegen build script defined in `common/src/codegen.rs`.

A similar process is handled for ABI generates files. They are moved into the local `target/` folder as well, and a main `src/abi.rs` file is used to specify the path target.

To use the changes, ensure proto files include a fully qualified package name, e.g. `package example.type.v1`. I have updated the folder structures to use a versioned hierarchy for the files, but the only thing necessary is that the substreams codegen has the correct imports to the source pb files it needs. The proposed codegen does not rely on the folder structure of the proto files, only the files generated by the substreams codegen found within the `src/pb/` folder, which are subsequently moved to target.

Additionally, this PR updates substreams manifest to use the workspace targets for generated `.wasm` spkgs.


Ryanmtate:
resolves #https://github.com/messari/substreams/issues/20

0xbe1:
@Ryanmtate let me review after rebase, thanks!!!

0xbe1 (approved):
LGTM!!!

common/src/codegen.rs
```diff
@@ -0,0 +1,146 @@
...
+    let out_dir = out_dir.unwrap_or(DEFAULT_OUTPUT_DIR);
+    let pb_dir = Path::new("src").join("pb");
+    let target_pb_dir = Path::new(out_dir).join("pb");
+
+    // Remove previous generated files
+    fs::remove_dir_all(&target_pb_dir).ok();
+    fs::remove_dir_all(&pb_dir).ok();
+
+    // generate pb files under src/pb
+    Command::new("make")
```
this-username-is-taken, line 64: Would it be better to run this directly?
```
substreams protogen ./substreams.yaml --exclude-paths="sf/ethereum,sf/substreams,google"
```
We are not enforcing that `make codegen` will run the command above so there is a chance that someone modifies `make codegen` and build fails.

Alternatively I guess we can make a comment here saying that it's expected that every substream have a Makefile with `make codegen` and it's expected to run `substreams protogen`

common/src/codegen.rs
```diff
@@ -0,0 +1,146 @@
...
+    let pb_mod_content = pb_files
+        .iter()
+        .map(|(pb, _)| format!("pub mod {};", pb))
+        .collect::<Vec<_>>()
+        .join("\n");
+
+    fs::write("src/pb.rs", pb_mod_content).unwrap();
+
+    // Create target directories
+    fs::create_dir_all(&target_pb_dir).ok();
```
this-username-is-taken, line 96: is this necessary? We already created this on line 70

common/src/codegen.rs
```diff
@@ -0,0 +1,146 @@
...
+
+    // generate pb files under src/pb
+    Command::new("make")
+        .args(&["codegen"])
+        .status()
+        .expect("failed to codegen");
+
+    // Create target directories
+    fs::create_dir_all(&target_pb_dir).ok();
+    fs::create_dir("./src/pb/").ok();
```
this-username-is-taken, line 71: why not use `pb_dir` here?

common/src/codegen.rs
```diff
@@ -0,0 +1,146 @@
...
+        // let package_name = filename[0];
+        let name = filename[1];
+        let version = filename[2];
+        pb_files
+            .entry(name)
+            .or_insert(HashSet::new())
+            .insert(version.to_owned());
+    }
+
+    let pb_mod_content = pb_files
```
this-username-is-taken, line 87: maybe we can add a comment like:

```// @generated```

at the top of the file to make it clear that this is generated?

File: common/src/codegen.rs [hunk 5]
```diff
@@ -0,0 +1,146 @@
...
+    fs::create_dir_all(&target_pb_dir).ok();
+
+    // Move to target folders
+    fs::rename(&pb_dir, &target_pb_dir).ok();
+
+    fs::create_dir(pb_dir.clone()).ok();
+
+    for (filename, versions) in pb_files.iter() {
+        let pb_file = format!("src/pb/{}.rs", filename);
+        let content = versions
```
this-username-is-taken, line 105: maybe we can add a comment like:

```// @generated```

at the top of the file to make it clear that this is generated?

File: erc20-price/build.rs [hunk 6]
```diff
@@ -1,66 +1,10 @@
...
-    fs::write(
-        "src/abi/mod.rs",
-        abi_filenames
-            .iter()
-            .map(|contract| format!("pub mod {};", contract))
-            .collect::<Vec<_>>()
-            .join("\n"),
-    )
-    .unwrap();
+    // println!("cargo:rerun-if-changed=proto");
```
this-username-is-taken, line 5: is this left in here intentionally? what does it do?

this-username-is-taken (commented):
Left a couple minor comments. Great work overall!

PR #63, opened 2022-09-22
feature(integrations): add an integrations test harness for substreams modules

### LINKED ISSUE #62: Integration Test Harness
Create integration test harness for packages substream modules.

It should:

[ ] Load substreams spkg module and use the grpc sf endpoint to test the consumption of the stream;
[ ] Cache the results of the stream;
[ ] Check runs against cached results;
[ ] Save cached results to target folder (used by github actions);

Ryanmtate:
Resolves #62 

The integrations crate provides a module test harness, calling the `substreams` cli to run each of the modules and cache the results to be compared on the next run.

Cached results are saved in the `integrations/target/` folder and can be run locally. The cached results are not committed to the repository, instead they can be restored in the github ci workflow for comparison.

NOTE: There is some beginning work on calling the substreams grpc `blocks` endpoint directly using the substreams playground [example](https://github.com/streamingfast/substreams-playground/blob/master/consumers/rust/src/main.rs) as a guide, however, I have not been able to call the endpoint successfully and reverting to using the substreams CLI in the meantime.


0xbe1:
> There is some beginning work on calling the substreams grpc blocks endpoint directly using the substreams playground [example](https://github.com/streamingfast/substreams-playground/blob/master/consumers/rust/src/main.rs) as a guide, however, I have not been able to call the endpoint successfully and reverting to using the substreams CLI in the meantime.

let's just use substreams CLI

.github/workflows/ci.yaml
```diff
@@ -20,7 +22,16 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       - name: install buf
         uses: bufbuild/buf-setup-action@v1.7.0
-      - name: install rust-toolchain
+      - name: install protoc
```
0xbe1, line 25: how about using this action instead?
https://github.com/marketplace/actions/setup-protoc

Ryanmtate:
After talking with @this-username-is-taken today, need to give integrations further thought. Would like to compare the output of the subgraph against the output of the substreams module. Need to do some investigations into the best way to automate a subgraph <> substreams output comparison.

For now, the integration harness is just checking the output from a single module against subsequent runs. When changes to the schema are made it will change the result and fail tests. Therefore, there needs to be a version of the module associated with tests. Since schemas typically only have major versions, we may wish to adopt a package version that is tightly coupled with the versioning of the protobuf package.

PR #66, opened 2022-09-22
refactor(substreams-helper): rename price to erc20_price

0xbe1:


Ryanmtate:
Let's merge in #60 as it will have some conflicts with this.

PR #67, opened 2022-09-22
wip: rename map_price to map_eth_price runtime error

0xbe1:
To reproduce:
1. cd erc20-price
2. make build && make example

Error
```
Error: rpc error: code = Internal desc = unexpected termination: process block failed: panic at block 13000000: runtime error: invalid memory address or nil pointer dereference
```


0xbe1:
sf server error, fixed by eduard

PR #78, opened 2022-09-27
feat(substreams-helper): add curve calculations to price lib

### LINKED ISSUE #34: add curve calculations to price lib

0xbe1:
close #34


substreams-helper/src/price.rs
```diff
@@ -9,6 +9,7 @@ const CONFIG: Config = Config {
     ethereum: NetworkConfig {
         yearn_lens_oracle: hex!("83d95e0d5f402511db06817aff3f9ea88224b030"),
         chainlink_feed_registry: hex!("47fb2585d2c56fe188d0e6ec628a38b74fceeedf"),
+        curve_calculations: hex!("25BF7b72815476Dd515044F9650Bf79bAd0Df655"),
```
this-username-is-taken, line 12: @Ryanmtate do you know if there is a linter we could use to convert all hex to lowercase?

Ryanmtate: Maybe regex pattern match?

PR #86, opened 2022-09-30
network: add initial network proto and substreams module

Ryanmtate:
WIP. Adds initial network protobuf based on subgraph graphql types, see: https://github.com/messari/subgraphs/blob/master/subgraphs/network/schema.graphql


this-username-is-taken:
Not sure how much work it is but might be helpful to have something we can use to bootstrap new substreams. Something similar to `graph init` in `graph-cli`: https://github.com/graphprotocol/graph-cli. I will add this in the backlog

network/proto/v1/network.proto
```diff
@@ -0,0 +1,259 @@
...
+    // #### Quantitative Data ####
+  
+    // Number of cumulative unique authors (see type Author for explanation)
+    int64 cumulative_unique_authors = 5;
+  
+    // The total number of blocks in the blockchain
+    int64 block_height = 6;
+  
+    // Total difficulty to reach the current block
+    int64 cumulative_difficulty= 7;
```
this-username-is-taken, line 28: minor but might be helpful to have a linter to catch this kind of thing: `cumulative_difficulty=`

Ryanmtate:
> Not sure how much work it is but might be helpful to have something we can use to bootstrap new substreams. Something similar to `graph init` in `graph-cli`: https://github.com/graphprotocol/graph-cli. I will add this in the backlog

Opened #88 to track

network/proto/v1/network.proto
```diff
@@ -0,0 +1,370 @@
...
+  
+    // #### Snapshots ####
+  
+    // Daily metrics for this blockchain
+    DailySnapshots daily_snapshots = 16;
+  
+    // Hourly metrics for this blockchain
+    HourlySnapshots hourly_snapshots = 17;
+
+    NetworkSnapshot network_snapshot = 18;
```
Ryanmtate, line 64: Adding `NetworkSnapshot` based on discussion on https://www.notion.so/messari/Subgraph-Schema-feedback-bc070812d29e4aea9886e4d7201ef206

These `Stats` are in addition to the cumulative values in the Network schema.

Q: Is this the way we are thinking about the `Network` schema?

Ryanmtate:
IMHO, the number of stats that we're including in the `Network` schema might be better handled as individual map method, as opposed to trying to compute the entire stats on read. I have a feeling that the computation will be slow and will cause indexing to be slow.

Probably worth re-visiting the schema and breaking the stats into stand-alone types that allow users to pick and choose which stats they care most about. having a singular `Network` schema to aggregate all the information together seems overloaded.

For example, another approach might be just a map for the cumulative values and keep the `Stats` in separate map methods.

I'm also a little unsure how parallelism will be handled for storing values to handle the `Stats` properties. 

There is no data types for a list/array/vector in substreams, or a `push` merge strategy that would allow you add a new item to a list. It would seem we will need a message type to series a vector of values, e.g.

```proto
message Values {
repeated BigInt values = 1;
}

```

This would be fine, however, we cannot read AND write to the same store in the same method, nor can we have circular dependencies, which begs the question how do we keep a list of values to perform operations-- e.g. the `values` property of `Stats` schema?

Open to suggestions, but I think it would almost be better to read the network into a SQL db that can have stored procedures to run various kinds of analytics.

I'm not sure if substreams is the best solution to handle these quantitative types of queries.

network/proto/v1/network.proto
Ryanmtate: @dmelotik Thoughts on the schema? I also added a note below discussing some of the limitations we may encounter with substreams in terms of how we handle calculations, e.g. keeping a record of values in the `Stats` schema.

this-username-is-taken:
merging for hand-over

PR #92, opened 2022-10-14
Fix and refactor

robinbernon:
Fixing all compiler errors and warnings. Should now be able to clone this repo and run "make test" without any issues. For this to work a change was required to the substreams client: https://github.com/streamingfast/substreams/pull/53.


.github/workflows/ci.yaml
```diff
@@ -14,7 +14,7 @@ jobs:
       - name: install substreams cli
         uses: jaxxstorm/action-install-gh-release@v1.7.1
         with:
-          repo: streamingfast/substreams
+          repo: robinbernon/substreams-messari
```
this-username-is-taken, line 17: why `robinbernon/substreams-messari`?

Makefile
```diff
@@ -8,6 +8,14 @@ build-all:
 	$(MAKE) -C erc721 build
 	$(MAKE) -C compound-v2 build
 
-.PHONY: erc20-price-spkg
-erc20-price-spkg:
-	cd erc20-price && make pack && mv erc20-price-substreams-v0.1.0.spkg ../target
\ No newline at end of file
+.PHONY: example-all
```
this-username-is-taken, line 11: What are your thoughts on changing this to `run-all` and changing `example` to `run`? I feel like the latter is more self-explanatory

this-username-is-taken (dismissed):
Great work! Added a couple minor comments otherwise LGTM.

Makefile
robinbernon: Yep sounds good! - Would you like me to change all individual `make example` commands to `make run` also?

.github/workflows/ci.yaml
robinbernon: They have now merge my PR into their main branch so I've put this back to what it was originally. Unfortunately they have not yet made a release for the new commit to master so the build agent is still pulling down the previous version which is why build is still failing.I'll keep an eye out for when they make the new release.

Makefile
this-username-is-taken: yes please

this-username-is-taken:
btw, we have some rough guidelines around PR naming convention for the subgraph repo: https://github.com/messari/subgraphs/blob/master/docs/Contributing.md#naming-conventions.

Doesn't 100% applies here but you can use it (and the previous PRs' name) as a reference.

.github/workflows/ci.yaml
this-username-is-taken: sounds good!

this-username-is-taken (approved):
Great work!

PR #95, opened 2022-10-25
Adding a cmd line client to streamline all non business logic operations

robinbernon:


Cargo.toml
```diff
@@ -7,12 +7,16 @@ members = [
...
 ]
+exclude = ["messari-cli"]
 
 [workspace.dependencies]
 substreams = "~0.0.20"
-substreams-ethereum = "0.2.2-rc.2"
+-substreams-ethereum = "0.2.2-rc.2"
+
+[build]
+target = "wasm32-unknown-unknown"
```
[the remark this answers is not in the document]
robinbernon: This is just moving the build requirement from .cargo/config.toml to ./Cargo.toml. This is due to all projects within the outer repo inheriting the build requirements from .cargo/config.toml - even the projects that are excluded from the workspace like messari-cli. Once the build target is moved to ./Cargo.toml it allows projects outside the workspace like messari-cli to be able to build with default targets rather than the workspace target.

messari-cli/src/abi.rs
```diff
@@ -0,0 +1,356 @@
...
+    project_dir: &PathBuf,
+    is_add_operation: bool,
+) {
+    let (abi_infos, contract_names) = abis_arg.get_abi_info(&protocol_and_network_info.protocol);
+
+    if abi_infos.is_empty() {
+        let abi_string = if is_add_operation {
+            get_input("Abi", None, false)
+        } else {
+            let abi_string = get_input("Abi (Leave blank to skip)", Some("Abi"), true);
```
this-username-is-taken, line 117: I'd change `Abi` to `Smart Contract Address (to fetch ABI)`, which makes it more clear on what input you are asking for

robinbernon: I have it as ABI because the ABI can be sourced from both smart contracts but also locally. So as it doesn't just represent a contract address I thought it would be best to keep it abi - as the umbrella term for both smart contract addresses and local file paths.

this-username-is-taken: I see. In that case, would be helpful to add some instructions around that. It wasn't clear that you could do that just by interacting with the CLI

robinbernon: There are. Run any of the two commands with the --help flag and it will tell you that you can enter either a local path or a contract address

messari-cli/src/file_modification/substreams_yaml.rs
```diff
@@ -0,0 +1,69 @@
...
+pub(crate) struct SubstreamsYaml {
+    substreams_yaml_dir: PathBuf, // This is parent directory for the substreams.yaml file
+    yaml: Yaml,
+}
+
+impl SubstreamsYaml {
+    pub(crate) fn new(project_name: &str, substreams_filepath: &PathBuf) -> Self {
+        let substreams_yaml_dir = substreams_filepath.parent().unwrap().to_path_buf();
+
+        let yaml_contents = format!(
```
this-username-is-taken, line 14: For this kind of codegen where the content is mostly the same and we are just substituting some variables, I think a declarative approach is much more readable and maintainable. We use Mustache for subgraph templating, seems like it also supports Rust: https://github.com/nickel-org/rust-mustache

this-username-is-taken (dismissed):
Great work! Can't wait to use it.

Issue #97, opened 2022-11-02
Integrate Chainlink prices using events (from aggregator contracts)


this-username-is-taken:
An example would be listening for the `AnswerUpdated` events emitted by `AccessControlledAggregators` (e.g. https://etherscan.io/address/0xd866a07dea5ee3c093e21d33660b5579c21f140b)

Subgraph example:
https://github.com/OpenPredict/chainlink-prices-subgraph/blob/master/src/mapping.ts

Resources:
- https://docs.chain.link/docs/data-feeds/price-feeds/
- https://talk.linkpool.io/t/the-chainlink-price-feed-contract-stack/44

this-username-is-taken:
see https://github.com/messari/substreams/pull/106

PR #98, opened 2022-11-02
feat(erc20-price): get price from chainlink feeds

this-username-is-taken:
See issue https://github.com/messari/substreams/issues/97


this-username-is-taken:
see https://github.com/messari/substreams/pull/106

PR #101, opened 2022-11-04
Uniswap V2; Add entities to dex proto and started indexing them. Also, indexing pool counts. 

steegecs:
Added some entities to the dex_amm.v1 proto. These were swaps, deposits, withdraws, and pools. Created mapping modules for each (pools are very incomplete). 

Also, created a module for storing counts which we would like to generalize.


uniswap-v2/src/pb/messari.common.v1.rs
```diff
@@ -0,0 +1,91 @@
+// @generated
```
this-username-is-taken, line 1: you shouldn't check in generated files

uniswap-v2/substreams_0.0.21_linux_x86_64/LICENSE
```diff
@@ -0,0 +1,202 @@
+
```
this-username-is-taken, line 1: why is this checked in?

common/proto/dex_amm.proto
```diff
@@ -6,22 +6,135 @@ import "erc20.proto";
...
-  string name = 1;
-  string address = 2; // bytes: Address
-  repeated messari.erc20.v1.ERC20Token input_tokens = 3;
-
-  // Metrics
-  string total_value_locked = 100; // string: BigDecimal
-}
-
-message PoolToken {
   string address = 1; // bytes: Address
```
this-username-is-taken, line 9: any particular reason for flattening the `PoolToken` struct into `Pool`?

common/proto/dex_amm.proto
```diff
@@ -6,22 +6,135 @@ import "erc20.proto";
 package messari.dex_amm.v1;
 
 message Pool {
```
this-username-is-taken, line 8: I think one way to design this is to have a common `Pool` struct across all protocols, with fields like TVL. Each protocol type would extend the `Pool` struct to include more protocol-specific fields. This way we can have shared methods like `SumTVL` that would take an array of `Pools` and return the total protocol TVL, and have it be applicable for all protocols

common/proto/dex_amm.proto
```diff
@@ -6,22 +6,135 @@ import "erc20.proto";
...
+  string pool_address = 11;
+}
+
+message Deposits {
+  repeated Deposit items = 1;
+}
+
+message Withdraw {
+  // Transaction hash of the transaction that emitted this event
+  string tx_hash = 1;
```
this-username-is-taken, line 65: I've been thinking about creating a container `Transaction` struct that would include fields like `tx_hash`, `block_number`, and `timestamp` so we don't need to repeat them for each message. What are your thoughts?

uniswap-v2/src/lib.rs
```diff
@@ -192,16 +201,255 @@ fn map_swap_events(
...
+        s.add(event.log_ordinal, &keyer::usage_count_key(&"Swap".to_string(), &"Day".to_string(), &day_id), 1);
+        s.add(event.log_ordinal, &keyer::usage_count_key(&"Swap".to_string(), &"Hour".to_string(), &hour_id), 1);
+        s.add (event.log_ordinal, &keyer::transaction_count_key(&"Day".to_string(), &day_id), 1);
+        s.add (event.log_ordinal, &keyer::transaction_count_key(&"Hour".to_string(), &hour_id), 1);
+    }
+}
+
+#[substreams::handlers::store]
+fn map_volumes_from_swaps(swaps: dex_amm::Swaps, s: store::StoreAddBigInt) {
+    for swap in swaps.items {
```
this-username-is-taken, line 405: is it possible to refactor this into `substreams_helper`?

uniswap-v2/src/lib.rs
```diff
@@ -192,16 +201,255 @@ fn map_swap_events(
...
+    for event in swaps.items {
         let pool_key = keyer::swap_key(&event.tx_hash, event.log_index);
         output.set(0, &pool_key, &proto::encode(&event).unwrap());
     }
 }
 
 #[substreams::handlers::store]
-fn store_swap_count(swap_events: uniswap::SwapEvents, s: store::StoreAddInt64) {
-    s.add(0, "total", swap_events.items.len() as i64);
+fn store_counts(clock: Clock, pair_created_events: uniswap::PairCreatedEvents, mint_events: uniswap::MintEvents, burn_events: uniswap::BurnEvents, swap_events: uniswap::SwapEvents, s: store::StoreAddInt64) {
```
this-username-is-taken, line 376: I wonder if there is a more generic way of implementing this (e.g. have a way to accumulate count for any repeated messages). @robinbernon do you have any thoughts on this?

uniswap-v2/src/pb/messari.common.v1.rs
steegecs: Yes - my bad! Cleaned up.

common/proto/dex_amm.proto
steegecs: Yeah! I hadn't had time to push the changes yet, but I have done something like this. I just chose the name `UsageEvents`, but it can be changed to whatever seems most appropriate. I did not use transaction, because multiple of these events can occur with a single transaction.

uniswap-v2/src/lib.rs
steegecs: Yes - definitely.

common/proto/dex_amm.proto
steegecs: Actually, using `Transaction` as an abstraction above `UsageEvents` might be a good idea. Then we can more easily keep track of transaction counts in the `UsageMetricsDailySnapshots` as well.

uniswap-v2/src/lib.rs
steegecs: I just pushed an update that generalizes it a little more. Another option would be to track the counts as a field in the `Transaction` or `UsageEvents` proto message. 

More ideas welcome!

common/proto/dex_amm.proto
steegecs: Yeah I like that idea, I think we should do it.

common/proto/dex_amm.proto
steegecs: It would be cleaner to build this data structure if we mapped all the raw event data together in a single mapping module instead of `map_swap_events`, `map_deposit_events`, etc, because you would then have to group the events back together by transaction hash. But it trades off with not being able to immediately work with specific raw event data in isolation, since you will have to first reprocess the `Transaction` message to get back out the event data that you want to work with.

steegecs: Here is an option mapping raw events to the `Transaction` and `Event` abstractions. 
Screen Shot 2022-11-06 at 8 23 44 PM

Or an option using a similar scheme, but for the final `Deposit`, `Withdraw`, and `Swap` messages.
Screen Shot 2022-11-06 at 8 30 13 PM

this-username-is-taken: what's the benefit of grouping the event type with the events themselves?

this-username-is-taken: what's the benefit of grouping the event type with the events themselves?

steegecs: Its not just the type. The `type` field allow you to include the event specific data, so the Event message can contain `swap`, `deposit`, or `withdraw` event data in addition to the data common to events, and is then identifiable as being from one of those categories. So I can pass a list of `Event` messages to a mapping or store module and handle each depending on the event type like here: https://github.com/messari/substreams/blob/steegecs/develop/uniswap-v2/src/lib.rs#:~:text=pub%20fn-,map_pool_deltas,-(events%3A.

steegecs: Though, if we wanted to use a pattern like this with the `Event` message but more general, we could serialize the event data to a string and deserialize it within a module, but I'd think we'd still want another piece of data to be able to identify the `type` or `name` of the event. .

steegecs: > v2/src/lib.rs#:~:text=%3A%3Amap%5D-,pub%20fn%20map_pool_deltas(events%3A%20dex_amm%3A%3AUsageEvents)%20%2D%3E,%7D,-%23%5Bsubstreams.

uniswap-v2/src/lib.rs
```diff
@@ -192,16 +200,265 @@ fn map_swap_events(
...
+    for swap in swaps.items {
+        for (token_index, token_address) in swap.tokens_in.iter().enumerate() {
+            let amount_bi = utils::convert_string_to_bigint(&swap.amounts_in[token_index]);
+            s.add(0, &keyer::pool_input_token_amounts_key(&swap.pool_address, &token_address), &amount_bi);
+        }
+    }
+}
+
+#[substreams::handlers::map]
+pub fn map_pool_deltas(events: dex_amm::UsageEvents) -> Result<uniswap::PoolTokenTvlDeltas, substreams::errors::Error> {
```
this-username-is-taken, line 385: what are pool deltas used for?

PR #105, opened 2022-11-14
Messari cli extensions

robinbernon:
Main changes:
  - CLI now produces new projects that can immediately be run using the "make run" command. (also added to make run-all cmd)
  - abi and proto files are now generated in a consistent manor avoiding reshuffling
  - Removed "make codegen" cmds from Makefiles (incorporated cmd into common::codegen() fn)
  - Fixed bug  for imaginary blank file change errors in git for windows OS
  - Switched crate dependency declaration from wasm32 only target to general target to allow for quicker crate additions using cargo-edit


messari-cli/src/commands/init.rs
```diff
@@ -196,6 +201,58 @@ fn create_substreams_project(
     let mut project_makefile_contents = MakeFile::new(&project_makefile);
     project_makefile_contents.add_build_operation();
 
+    let mut yaml_contents = SubstreamsYaml::new(project_name.as_str(), &substreams_yaml);
```
this-username-is-taken, line 204: I provided this feedback in the previous PR but wanted to ask again here: what's the reasoning on using a procedural approach instead of a declarative approach here? It seems like this is growing in complexity which will make it more and more difficult to maintain. Also, we now have a number of static files that are copied as templates, it would make sense for all files to be generated from templates for consistency.

messari-cli/src/commands/init.rs
```diff
@@ -196,6 +201,58 @@ fn create_substreams_project(
...
+            None,
+            UpdatePolicy::Set,
+            "proto:messari.erc20.v1.TransferEvents".to_string(),
+            vec![Input {
+                input_type: InputType::Map,
+                input_value: "map_example".to_string(),
+            }],
+        ));
+        yaml_contents.add_protobuf_files(vec![get_repo_root_folder()
+            .join("common")
```
this-username-is-taken, line 240: `common` and `proto` are not needed

common/src/codegen.rs
```diff
@@ -46,70 +48,126 @@ pub fn generate_abi(out_dir: Option<&str>) -> Result<(), Error> {
...
+            })
+            .collect::<Vec<_>>();
+
+        pb_files_vec.sort_by(|(filename1, _), (filename2, _)| filename1.cmp(filename2));
+        pb_files_vec
+    };
+
     if !target_pb_dir.exists() {
-        fs::create_dir_all(&target_pb_dir).unwrap();
+        fs::create_dir(&target_pb_dir).unwrap();
```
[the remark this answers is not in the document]
robinbernon: Have added a comment for why this is possible in commit: 47336a5a238ea690962bed5a7feb256d98a5b4d4

this-username-is-taken (approved):

messari-cli/src/commands/init.rs
robinbernon: As discussed today and yesterday in person the idea here is that we want procedural generation to allow us to also use the code for mutating existing documents when adding our extra data. This will be important later on especially for when we start to do any extra code generation around the modification of map or store methods, including also the changes based around protos which also need to be reflected in the yaml document. The reason for the inconsistency with the template files is that, in particular, the two files erc20.json and lib.rs won't be templates for long. Soon it would be good to create them procedurally based on user requirements to give a greater setup leading to less work for the developer.

PR #107, opened 2022-11-29
feat(erc20-holdings): Add `store_balance_usd` module 

harsh9200:


erc20-holdings/src/lib.rs
```diff
@@ -119,3 +125,43 @@ fn store_balance(transfers: erc20::TransferEvents, output: store::StoreAddBigInt
...
+    output: store::StoreSetBigDecimal,
+) {
+    for transfer in transfers.items {
+        let token_price =
+            match prices.get_last(format!("chainlink_price:{}", &transfer.token_address)) {
+                Some(x) => BigDecimal::from_str(x.price_usd.as_str()).unwrap(),
+                None => BigDecimal::from_str("0").unwrap(),
+            };
+
+        let token_denominator = abi::erc20::functions::Decimals {}
```
this-username-is-taken, line 143: why not get it from `Erc20Price`? Contract calls are orders of magnitude slower and should be avoided at all costs

erc20-holdings/src/lib.rs
```diff
@@ -119,3 +125,43 @@ fn store_balance(transfers: erc20::TransferEvents, output: store::StoreAddBigInt
...
+#[substreams::handlers::store]
+fn store_balance_usd(
+    transfers: erc20::TransferEvents,
+    balances: store::StoreGetBigInt,
+    prices: StoreGetProto<Erc20Price>,
+    output: store::StoreSetBigDecimal,
+) {
+    for transfer in transfers.items {
+        let token_price =
+            match prices.get_last(format!("chainlink_price:{}", &transfer.token_address)) {
```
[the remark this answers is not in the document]
harsh9200: yeah, that's a good idea.

erc20-price/proto/v1/erc20_price.proto
```diff
@@ -2,10 +2,17 @@ syntax = "proto3";
...
 
+message Token {
+  string address = 1;
+  string symbol = 2;
+  uint64 decimals = 3;
+}
+
 message Erc20Price {
-  string token_address = 1;
+  Token token = 1;
```
this-username-is-taken, line 12: why not just use `Erc20Token`?
https://github.com/messari/substreams/blob/90e50e21abd5288f720aaf3d8f3feeba3a4d01bf/common/proto/erc20.proto#L9

erc20-price/proto/v1/chainlink.proto
```diff
@@ -5,9 +5,14 @@ package messari.chainlink.v1;
...
   string description = 2;
+
   string base = 3;
   string base_address = 4;
-  string quote = 5;
-  string quote_address = 6;
-  uint64 decimals = 7;
+  uint64 base_decimals = 5;
+  
+  string quote = 6;
```
this-username-is-taken, line 13: we can do the same here (e.g. `Erc20Token quote_asset`)

erc20-price/proto/v1/erc20_price.proto
harsh9200: Yeah, good point. Let me do that

erc20-price/src/utils.rs
```diff
@@ -1,151 +1,180 @@
...
+            ("DATA", "33d63ba1e57e54779f7ddaeaa7109349344cf5f1"),
+            ("FORTH", "77fba179c79de5b7653f68b5039af940ada60ce0"),
+            ("XCN", "a2cd3d43c775978a96bdbf12d733d5a1ed94fb18"),
         ]);
 
         token_mapping
     };
 }
+
+pub fn get_erc20_token(token_address: String) -> Option<Erc20Token> {
```
this-username-is-taken, line 160: why not use `substreams_helper::erc20::get_erc20_token` instead?

erc20-price/src/lib.rs
```diff
@@ -111,16 +114,14 @@ fn store_chainlink_aggregator(block: eth::Block, output: StoreSetProto<Aggregato
             let aggregator = Aggregator {
                 address: Hex(&aggregator_address).to_string(),
                 description: description.clone(),
-                base: base_quote[0].to_string(),
-                base_address: base_address.clone(),
-                quote: base_quote[1].to_string(),
-                quote_address: quote_address.clone(),
+                base_asset: utils::get_erc20_token(base_address),
```
this-username-is-taken, line 117: it's gonna be very expensive calling this every time. Since this data doesn't change, we should have a separate module to retrieve this data so we don't need to query it over and over again

PR #108, opened 2022-12-01
feat(erc20-price): Add Uniswap Price Feeds

harsh9200:


erc20-price/src/lib.rs
```diff
@@ -188,3 +190,157 @@ fn store_chainlink_price(
...
+                let reserve0 = event
+                    .reserve0
+                    .to_decimal(pair.token0.clone().unwrap().decimals);
+                let reserve1 = event
+                    .reserve1
+                    .to_decimal(pair.token1.clone().unwrap().decimals);
+
+                // TODO: Add a check for mininmum liquidity threshhold.
+
+                match pair.token0.clone().unwrap().symbol.as_str() {
```
this-username-is-taken, line 251: We need to use address and not symbol. Symbols are not unique identifiers

harsh9200: yeah, my bad. I will change this to address.

erc20-price/src/lib.rs
```diff
@@ -188,3 +190,157 @@ fn store_chainlink_price(
...
+                    symbol if ["DAI", "USDC", "USDT"].contains(&symbol) => {
+                        let token_price = reserve0.clone() / reserve1.clone();
+
+                        let erc20price = Erc20Price {
+                            token: pair.token1.clone(),
+                            price_usd: format!("{:.7}", token_price.to_string()),
+                            block_number: block.number,
+                            source: 2,
```
this-username-is-taken, line 259: this should be an enum, can put into `substreams-helper`

erc20-price/src/lib.rs
```diff
@@ -188,3 +190,157 @@ fn store_chainlink_price(
...
+                        };
+
+                        output.set(
+                            0,
+                            keyer::uniswap_asset_key(&pair.token0.clone().unwrap().address),
```
this-username-is-taken, line 291: i don't think you can assume the key to be always `token0`, you will need a check here

erc20-price/src/lib.rs
```diff
@@ -188,3 +190,157 @@ fn store_chainlink_price(
         }
     }
 }
+
+#[substreams::handlers::store]
+fn store_pair_created_events(block: eth::Block, output: StoreSetProto<PairCreatedEvent>) {
+    for log in block.logs() {
+        if let Some(event) = factory::events::PairCreated::match_and_decode(log) {
```
this-username-is-taken, line 197: I wonder if this catches ALL uniswap forks instead of just uniswap. Can you investigate?

erc20-price/src/lib.rs
harsh9200: yeah, that's a typo. The last output set should be using `token1`.

erc20-price/src/lib.rs
this-username-is-taken: I meant `source: 2` should be an enum

erc20-price/src/lib.rs
```diff
@@ -188,3 +190,158 @@ fn store_chainlink_price(
...
+    output: StoreSetProto<Erc20Price>,
+) {
+    for log in block.logs() {
+        if let Some(event) = pair::events::Sync::match_and_decode(log) {
+            let pair_address = Hex(log.address()).to_string();
+
+            if let Some(pair) = store.get_last(keyer::pair_info_key(&pair_address)) {
+                let reserve0 = event
+                    .reserve0
+                    .to_decimal(pair.token0.clone().unwrap().decimals);
```
robinbernon, line 245: You can skip a clone by taking a reference of the object before unwrapping. Try doing something like: ```pair.token0.as_ref().unwrap().decimals``` instead.

erc20-price/src/lib.rs
```diff
@@ -188,3 +190,158 @@ fn store_chainlink_price(
...
+                let reserve1 = event
+                    .reserve1
+                    .to_decimal(pair.token1.clone().unwrap().decimals);
```
robinbernon, line 248: Same thing here as above

erc20-price/src/lib.rs
```diff
@@ -188,3 +190,158 @@ fn store_chainlink_price(
...
+
+                // TODO: Add a check for mininmum liquidity threshhold.
+
+                match pair.token0.clone().unwrap().address.as_str() {
```
robinbernon, line 252: Same thing here as above

erc20-price/src/lib.rs
```diff
@@ -188,3 +190,158 @@ fn store_chainlink_price(
...
+                            0,
+                            keyer::uniswap_asset_key(&pair.token0.clone().unwrap().address),
+                            &erc20price,
+                        );
+                    }
+                    address if utils::TOKENS.get("ETH").unwrap().eq(&address) => {
+                        let eth_price = match chainlink_prices.get_last(keyer::chainlink_asset_key(
+                            &pair.token0.clone().unwrap().address,
+                        )) {
+                            Some(price) => BigDecimal::from_str(price.price_usd.as_str()).unwrap(),
```
robinbernon, line 273: Think we should change the field ```string price_usd = 2;``` to use a BigDecimal field instead to avoid string parsing

erc20-price/src/lib.rs
```diff
@@ -188,3 +190,158 @@ fn store_chainlink_price(
...
+
+                        let erc20price = Erc20Price {
+                            token: pair.token1.clone(),
+                            price_usd: token_price.to_string(),
+                            block_number: block.number,
+                            source: 2,
+                        };
+
+                        output.set(
+                            0,
```
robinbernon, line 264: You are parsing logs here so you should use the log ordinal rather than just setting 0 here

erc20-price/src/lib.rs
```diff
@@ -188,3 +190,158 @@ fn store_chainlink_price(
...
+
+                        let erc20price = Erc20Price {
+                            token: pair.token0.clone(),
+                            price_usd: token_price.to_string(),
+                            block_number: block.number,
+                            source: 2,
+                        };
+
+                        output.set(
+                            0,
```
robinbernon, line 311: Same thing :)

erc20-price/src/lib.rs
```diff
@@ -188,3 +190,158 @@ fn store_chainlink_price(
...
+                            keyer::uniswap_asset_key(&pair.token0.clone().unwrap().address),
+                            &erc20price,
+                        );
+                    }
+                    _ => {}
+                }
+
+                match pair.token1.clone().unwrap().address.as_str() {
```
robinbernon, line 299: Take ref then unwrap to avoid clone

erc20-price/src/lib.rs
```diff
@@ -188,3 +190,158 @@ fn store_chainlink_price(
...
+
+                        output.set(
+                            0,
+                            keyer::uniswap_asset_key(&pair.token1.clone().unwrap().address),
+                            &erc20price,
+                        );
+                    }
+                    address if utils::TOKENS.get("ETH").unwrap().eq(&address) => {
+                        let eth_price = match chainlink_prices.get_last(keyer::chainlink_asset_key(
+                            &pair.token1.clone().unwrap().address,
```
robinbernon, line 318: Same thing :)

erc20-price/src/lib.rs
```diff
@@ -188,3 +190,158 @@ fn store_chainlink_price(
...
+                        let erc20price = Erc20Price {
+                            token: pair.token0.clone(),
+                            price_usd: token_price.to_string(),
+                            block_number: block.number,
+                            source: 2,
+                        };
+
+                        output.set(
+                            0,
+                            keyer::uniswap_asset_key(&pair.token1.clone().unwrap().address),
```
robinbernon, line 338: Same thing :)

erc20-price/src/lib.rs
harsh9200: Gotcha

erc20-price/src/lib.rs
harsh9200: yeah sounds good, I will change this field to BigDecimal type.

erc20-price/src/lib.rs
harsh9200: Gotcha

erc20-price/src/lib.rs
this-username-is-taken: @robinbernon what type would `BigDecimal` be?

erc20-price/src/lib.rs
harsh9200: yes, this works for all the Uniswap forks.

erc20-price/src/modules/1_store_chainlink_aggregator.rs
```diff
@@ -0,0 +1,101 @@
...
+            if nested_aggregator.is_empty().not() {
+                // In `AggregatorFacade` contracts, the aggregator contract is nested two times.
+                // Example: 0xb103ede8acd6f0c106b7a5772e9d24e34f5ebc2c
+
+                aggregator_address = nested_aggregator;
+            }
+
+            let base_asset = match utils::TOKENS.get(base_quote[0]) {
+                Some(base) => {
+                    substreams_helper::erc20::get_erc20_token(String::from(base.deref())).unwrap()
```
robinbernon, line 52: Use ```substreams_helper::erc20::get_erc20_token(base.to_string()).unwrap()``` instead

erc20-price/src/modules/1_store_chainlink_aggregator.rs
```diff
@@ -0,0 +1,101 @@
...
+                        "Cannot find token mapping for base: {}",
+                        base_quote[0].to_string()
+                    );
+                    continue;
+                }
+            };
+
+            let quote_asset = match utils::TOKENS.get(base_quote[1]) {
+                Some(quote) => {
+                    substreams_helper::erc20::get_erc20_token(String::from(quote.deref())).unwrap()
```
robinbernon, line 65: Same as above :)

erc20-price/src/modules/2_store_chainlink_price.rs
```diff
@@ -0,0 +1,53 @@
...
+                if ["USD", "DAI", "USDC", "USDT"]
+                    .contains(&aggregator.quote_asset.unwrap().symbol.as_str())
+                    .not()
+                {
+                    // TODO: add logic for handling `ETH` quote.
+                    continue;
+                }
+
+                let token_price = event.current.to_decimal(aggregator.decimals);
+                let token_address = aggregator.base_asset.clone().unwrap().address;
```
robinbernon, line 36: Take ref to avoid clone

erc20-price/src/modules/3_store_pair_created_events.rs
```diff
@@ -0,0 +1,46 @@
...
+use crate::keyer;
+use crate::pb::erc20::v1::Erc20Token;
+use crate::pb::uniswap::v1::PairCreatedEvent;
+
+#[substreams::handlers::store]
+fn store_pair_created_events(block: eth::Block, output: StoreSetProto<PairCreatedEvent>) {
+    for log in block.logs() {
+        if let Some(event) = factory::events::PairCreated::match_and_decode(log) {
+            let token0_asset =
+                substreams_helper::erc20::get_erc20_token(Hex(event.token0.clone()).to_string())
```
robinbernon, line 16: You can pass a reference here rather than cloning

erc20-price/src/modules/5_map_eth_price.rs
```diff
@@ -0,0 +1,48 @@
...
+            .unwrap();
+
+        prices.items.push(Erc20Price {
+            token: Some(Erc20Token {
+                address: token.address,
+                name: token.name,
+                symbol: token.symbol,
+                decimals: token.decimals,
+            }),
+            block_number: block_number,
```
robinbernon, line 40: You change this line from ```block_number: block_number,``` to ```block_number,```

erc20-price/src/modules/5_map_eth_price.rs
```diff
@@ -0,0 +1,48 @@
...
+) -> Result<Erc20Prices, substreams::errors::Error> {
+    let mut prices = Erc20Prices { items: vec![] };
+
+    for erc20_token in erc20_tokens {
+        let token_price = price::get_price(Network::Ethereum, block_number, erc20_token.clone())
+            .map_err(|e| {
+                substreams::errors::Error::Unexpected(format!("Failed to get price: {}", e))
+            })?;
+
+        let token = substreams_helper::erc20::get_erc20_token(Hex(erc20_token.clone()).to_string())
```
robinbernon, line 30: You can pass a reference here instead

erc20-price/src/lib.rs
this-username-is-taken: since this will be used by a lot of people, I'd rather be selective about which price feed to offer. I think the filtering should be happening inside of this module instead of outside

harsh9200: yeah, that makes sense. I will add a check to filter out prices only from known factory addresses.

this-username-is-taken:
Great work!

PR #109, opened 2022-12-05
Adding network substream to repo

robinbernon:
Adding network substream to repo.


network/proto/aggregate_data.proto
```diff
@@ -0,0 +1,33 @@
+syntax = "proto3";
+package aggregate_data;
+
+message BigInt {
+    bytes bytes = 1;
+}
+
+message PreCalculatedAggregates {
+    string timestamp = 1; // Timestamp for the given day/hour that this aggregated data is associated with
+    BigInt blocks = 2; // Only used for daily tracking and not hourly
```
this-username-is-taken, line 10: IIRC, `bytes` don't get displayed properly in the console (displayed as bytes and not human readable). That was the reason I used string instead. Any chance SF fixed this?

network/proto/v1/network.proto
```diff
@@ -41,16 +41,16 @@ message Network {
     // Total rewards emitted on the network in native asset
     BigInt cumulative_rewards = 11;
   
-    // Total number of transactions that have occured on this network
+    // Total number of transactions that have occurred on this network
     BigInt cumulative_transactions = 12;
   
     // Total size of the network in native units (e.g. Bytes in Arweave and Ethereum)
-    BigInt cumulative_size = 13;
+    BigInt cumulative_size  = 13;
```
this-username-is-taken, line 48: extra space here. Should we get a linter to catch these kind of things?

network/proto/aggregate_data.proto
robinbernon: Yeh good catch - will put this back to i64. I believe it should be readable

network/proto/v1/network.proto
robinbernon: I guess we could put this as a step in the workflow: https://github.com/bufbuild/buf-lint-action. We already need to install the buf cli so I guess it wouldn't take up much more time for the build stage that it is already taking

network/src/utils.rs
```diff
@@ -0,0 +1,60 @@
+use substreams::scalar::BigInt;
+use substreams_ethereum::pb::eth::v2::{self as eth};
+use substreams_ethereum::scalar::BigIntSign;
+use crate::pb::aggregate_data;
+
+/// Returns the timestamp for the start of the most recent day
+pub(crate) fn get_latest_day(timestamp: i64) -> i64 {
```
this-username-is-taken, line 7: can you move the functions here that may be used across different substreams into `substreams-helper`?

network/src/modules/1_unique_tracking.rs
```diff
@@ -0,0 +1,33 @@
...
+use substreams_ethereum::pb::eth::v2::{self as eth};
+use substreams::store;
+use substreams::store::{StoreSet, StoreSetIfNotExists};
+
+use crate::block_handler::BlockHandler;
+use crate::pb::aggregate_data::BigInt;
+use crate::utils::i64_to_str;
+
+#[substreams::handlers::store]
+pub fn unique_tracking(block: eth::Block, output: store::StoreSetI64) {
```
this-username-is-taken, line 11: which substreams module uses this function? Doesn't seem like this is referenced in `substreams.yaml`

network/src/modules/1_unique_tracking.rs
```diff
@@ -0,0 +1,33 @@
...
+        output.set(0, format!("t:{}", author), &1); // (Any number can be used here for differentiation as long as it's not set as blank..)
+        output.set(0, format!("d:{}:{}", i64_to_str(block_handler.days_timestamp()), author), &1);
+        output.set(0, format!("h:{}:{}", i64_to_str(block_handler.hours_timestamp()), author), &1);
+    }
+
+    output.set(0, "latest_day_timestamp", &block_handler.days_timestamp());
+    output.set(0, "latest_hour_timestamp", &block_handler.hours_timestamp());
+    output.set(0, "block_timestamp", &block_handler.timestamp());
+
+    // We gonna sneak in the gas limit here to save an extra map and store step... shh🤫 - don't tell anyone!
```
this-username-is-taken, line 31: 😂

network/src/pb/aggregate_data.rs
```diff
@@ -0,0 +1,233 @@
+// @generated
```
dmelotik: Why do we want to keep these auto-generated files and not ignore them?

network/proto/v1/network.proto
```diff
@@ -9,7 +9,7 @@ message Network {
     // Name of the network being indexed
```
dmelotik, line 9: Is the v1 folder our way of versioning protos (schemas)?

robinbernon: I think the aim is to version external schemas however I'd sayt that internal protos for data transfer between map and store methods don't need to follow this format. (I've got a commit that I'll push to the remote soon for handling non versioned binding generation that's still compatible with the versioned schemas)

network/src/modules/1_unique_tracking.rs
robinbernon: I realised that there is another piece of data that definitely needs it's own set store module to work so I'll put this into it as well

network/src/modules/1_unique_tracking.rs
robinbernon: Yep, just haven't pushed the config files to remote yet :)

network/src/utils.rs
robinbernon: Sure 👍

network/substreams.yaml
```diff
@@ -1,68 +1,97 @@
...
 
 binaries:
   default:
     type: wasm/rust-v1
-    file: ../target/wasm32-unknown-unknown/release/messari_networks_substreams.wasm
+    file: "../target/wasm32-unknown-unknown/release/network.wasm"
 
 modules:
-  - name: store_cumulative_values
+  - name: unique_tracking
```
dmelotik, line 15: Do we want to move away from prepending each module with `map_` or `store_`?

network/substreams.yaml
```diff
@@ -1,68 +1,97 @@
...
+    kind: store
+    initialBlock: 1
```
dmelotik, line 17: initial block 0?

network/substreams.yaml
robinbernon: I think as long as it's clear what the module is doing I think it's ok. I picked this for the module name as I thought this was the clearest thing to sum up what this module is doing, i.e., Tracking uniqueness among a few fields.

network/substreams.yaml
robinbernon: Block 0 is faulty. Not sure what is going on with it but it doesn't even have a timestamp associated with the block so I thought I'd just leave it out. (All other blocks seem to work fine so far)

network/substreams.yaml
dmelotik: sounds good, will keep that in mind during development

compound-v2/src/abi.rs
```diff
@@ -1,4 +1,3 @@
-// DO NOT EDIT - the file is generated by build script
```
dmelotik, line 1: out of scope of this PR

dmelotik:
@robinbernon can you add some module (inline) docs and/or a README? I think it will help readability and for future users. And it will (selfishly) help me understand what is going on better

network/src/block_handler.rs
```diff
@@ -0,0 +1,164 @@
...
+
+    pub fn gas_price(&self) -> BigInt {
+        if let Some(header) = self.0.header.as_ref() {
+            if let Some(base_fee_per_gas) = header.base_fee_per_gas.as_ref() {
+                return base_fee_per_gas.deserialize() * self.gas_used();
+            }
+        }
+
+        BigInt::zero()
+    }
```
jaimehgb, line 101: What should this return? The average price paid for each unit of gas used in the block?

Base fee only applies to blocks after eip1559, and doesn't necessarily reflect the price of the gas.

network/build/subgraph.yaml
dmelotik, line 1: @robinbernon is there a reason you have duplicates of `subgraph.yaml` and `schema.graphql`?

PR #112, opened 2022-12-08
Fixing pipeline and make pb imports to be done in only one file

robinbernon:
Made a temp fix for our workflow. (Have spoken to SF and they said they will make a patch for the next release so we can set the CI back to what is was previously for the next release.

Also got pb binding imports to be handled in just one file rather than a folder which was used previously.


.github/workflows/ci.yaml
```diff
@@ -43,4 +44,7 @@ jobs:
             target/
           key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
       - name: build all
-        run: make build-all
+        run: |
+          chmod a+x /opt/hostedtoolcache/streamingfast/substreams/v0.0.21/linux-x64/substreams_0.0.21_linux_x86_64/substreams
```
this-username-is-taken, line 48: any way we can avoid hardcoding this?

robinbernon: Don't think so. I have fixed the version for the substreams CLI to v21 so it definitely won't start breaking when they make the next release. And once we see they've made the first release with the fix to allow for proper unpacking of the binary during the installation stage we can revert the code back.

this-username-is-taken: gotcha, sounds good

this-username-is-taken (approved):
Looks great! Thanks for digging into the issue!

PR #113, opened 2022-12-09
feat(#major); ENS; Add ENS

harsh9200:
Added ENS substreams to index all ENS mappings and reverse mappings.


this-username-is-taken:
Now that Robin fixed the build, you should look into why it's failing. We should try to ensure every PR builds going forward.

ens/README.md
```diff
@@ -0,0 +1,3 @@
+# ENS
+
+NOTE: At the time of writing, the ens substream only works for ENS registered with ETH Base Registrar contract.
```
this-username-is-taken, line 3: You should add some description on what this module is for. Would also be helpful to add it into the PR description section

Cargo.toml
```diff
@@ -11,6 +11,7 @@ members = [
     "substreams-helper",
     "uniswap-v2",
     "solana-sample",
+    "ens",
```
this-username-is-taken, line 14: I'd rename this as `ens-lookup` as ENS can mean a lot of things (e.g. could be a substream for ENS the protocol)

ens-names/substreams.yaml
```diff
@@ -0,0 +1,45 @@
...
+    - ens.proto
+    - common.proto
+  importPaths:
+    - ./proto/v1
+    - ../common/proto
+
+binaries:
+  default:
+    type: wasm/rust-v1
+    file: "../target/wasm32-unknown-unknown/release/ens.wasm"
```
this-username-is-taken, line 20: should be `ens_names.wasm`

ens-names/proto/v1/ens.proto
```diff
@@ -0,0 +1,20 @@
...
+    string ens_name = 1;
+    string name_hash = 2;
+    
+    string label_name = 3;
+    string label_hash = 4;
+
+    string controller_address = 5;
+    string registrant_address = 6;
+
+    string transaction_hash = 7;
```
this-username-is-taken, line 15: can we also add `block_number` here?

ens-names/substreams.yaml
harsh9200: nice catch!

ens-names/proto/v1/ens.proto
harsh9200: Sure

this-username-is-taken:
what happens when the controller changes? do we update the store?

ens-names/src/keyer.rs
```diff
@@ -0,0 +1,7 @@
+pub fn ens_domain_key(asset_address: &String) -> String {
+    format!("domain:{}", asset_address)
```
this-username-is-taken, line 2: what does an `asset_address` refer to here?

ens-names/README.md
```diff
@@ -0,0 +1,3 @@
+# ENS Names
+
+NOTE: At the time of writing, the ens substream only works for ENS registered with ETH Base Registrar contract.
```
this-username-is-taken, line 3: you should add a more detailed description about what this substream does

harsh9200: Sure

ens-names/src/keyer.rs
harsh9200: I have renamed this to `ens_hash`.

this-username-is-taken (approved):
Looks great!

Issue #111, opened 2022-12-06
Substream to accumulate SOL balance for every account


this-username-is-taken:
you can use https://github.com/messari/substreams/tree/master/eth-balance as a reference

dmelotik:
@this-username-is-taken do we also want to be able to retrieve the balance of any address at any block?

It doesn't seem like the eth-balance supports that, but maybe I am misunderstanding or reading this wrong

this-username-is-taken:
yes, though we don't need to put data at every block in the store to retrieve that. SF provides a very nice way called "initial snapshots" that you can leverage. You can check it out here: https://github.com/this-username-is-taken/lido-balances, specifically https://github.com/this-username-is-taken/lido-balances/blob/37d00ff702e24588c0ae9e784fe93d5aa887ee57/main.py#L54.

You can also use this mode in `substreams run` by adding and `-i` option:
```
$ substreams run --help
Stream modules from a given package on a remote endpoint

Usage:
  substreams run <manifest> <module_name> [flags]

Flags:
  -h, --help                                 help for run
  -i, --initial-snapshots                    Load an initial snapshot at start block, before continuing processing.
  -k, --insecure                             Skip certificate validation on GRPC connection
  -o, --output string                        Output mode. Defaults to 'ui' when in a TTY is present, and 'json' otherwise
  -p, --plaintext                            Establish GRPC connection in plaintext
  -s, --start-block int                      Start block to stream from. Defaults to -1, which means the initialBlock of the first module you are streaming (default -1)
  -t, --stop-block string                    Stop block to end stream at, inclusively. (default "0")
      --substreams-api-token-envvar string   name of variable containing Substreams Authentication token (default "SUBSTREAMS_API_TOKEN")
  -e, --substreams-endpoint string           Substreams gRPC endpoint (default "api.streamingfast.io:443")
  ```

dmelotik:
Can use this tool to QA: https://messari.app.zettablock.com/my-workspace/queries/qu15-d679c88c-1a9b-4649-90a5-8f4614c16a8d

PR #114, opened 2022-12-10
feat(#major); add sol balance substream

dmelotik:
#111

### DIFF
--- a/sol-balance/rust-toolchain.toml
+++ b/sol-balance/rust-toolchain.toml
@@ -1,0 +1,4 @@
+[toolchain]
+channel = "1.64.0"
+components = [ "rustfmt" ]
+targets = [ "wasm32-unknown-unknown" ]
--- a/sol-balance/Cargo.toml
+++ b/sol-balance/Cargo.toml
@@ -1,0 +1,22 @@
+[package]
+name = "sol-balance"
+version = "0.1.0"
+description = "Messari's standardized substream for wallets sol balance"
+edition = "2021"
+repository = "https://github.com/messari/substreams"
+
+[lib]
+crate-type = ["cdylib"]
+
+[dependencies]
+prost = "0.11.2"
+ethabi = "17.2.0"
+num-bigint = "0.4"
+hex-literal = "0.3.4"
+substreams = { workspace = true }
+substreams-solana = { workspace = true }
+substreams-helper = { path = "../substreams-helper" }
+
+[build-dependencies]
+anyhow = "1"
+substreams-common = { path = "../common" }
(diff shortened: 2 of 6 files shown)


dmelotik:
Currently running into this issue: https://discord.com/channels/666749063386890256/982135810742697984/1038149086513205299

this-username-is-taken:
I believe I was able to workaround it by downgrading the version of the `substreams` package. See https://discord.com/channels/666749063386890256/982135810742697984/1038223878763335792

dmelotik:
@robinbernon would you mind taking a look at the last `TODO` comment. I am wondering if there is something wrong with my setup that it wants that `Vec` type

robinbernon:
> @robinbernon would you mind taking a look at the last `TODO` comment? I am wondering if there is something wrong with my setup that it wants that `Vec` type

No you are expected to be inputing a Vec<u8> type here. Vec<u8> is a generic placeholder for a list of bytes. You need to encode the struct to bytes in order to input it here. Try using the encode_to_vec() method to do this. (encode_to_vec() is a fn from the prost::Message trait which these structs all have derived already so you should just be able to call it directly without having to do any extra work here)

robinbernon:
Also, add this project to the makefile cmds in the root of the folder. It should be added to the build-all and run-all cmds.

dmelotik:
thanks @robinbernon

common/proto/sol_token.proto
```diff
@@ -0,0 +1,50 @@
...
+message Account {
+  string account_id = 1;
+  repeated Balance balances = 2;
+}
+
+message Accounts {
+  repeated Account items = 1;
+}
+
+message OuterString {
```
this-username-is-taken, line 48: What is this?

common/proto/sol_token.proto
```diff
@@ -0,0 +1,50 @@
...
+
+message Balance {
+  TokenAccount token = 1;
+  string balance = 2; // BigInt in token's native units
+  uint64 block_number = 3;
+  uint64 timestamp = 4;
+}
+
+message Account {
+  string account_id = 1;
```
this-username-is-taken, line 40: We should match the naming here with that of `evm-token`: https://github.com/messari/substreams/blob/d5faf2bbd021c26bc02b4abfa656a3dd391fedad/common/proto/evm_token.proto

common/proto/sol_token.proto
dmelotik: Removed, it was from printing the block as a string

sol-balance/rust-toolchain.toml
```diff
@@ -0,0 +1,4 @@
+[toolchain]
```
robinbernon, line 1: Same comments on this sorta stuff as with the other PRs :)

PR #117, opened 2022-12-17
feat(#major); Uniswap-v2; Add Uniswap-v2 substreams

harsh9200:


this-username-is-taken:
You should be consistent with Protobuf's naming conventions: https://developers.google.com/protocol-buffers/docs/style

common/proto/dex_amm.proto
```diff
@@ -5,23 +5,259 @@ import "erc20.proto";
 
 package messari.dex_amm.v1;
 
-message Pool {
-  string name = 1;
-  string address = 2; // bytes: Address
-  repeated messari.erc20.v1.ERC20Token input_tokens = 3;
+message Token {
```
this-username-is-taken, line 8: Why new `Token` type? We should building on existing types when possible.

this-username-is-taken:
Note that protobuf messages define the inputs and outputs of individual substream modules, and not that schema of the resulting subgraph. So you may want to rethink how these should be defined and organized instead of retrofitting them into our existing subgraph schema.

Below is the Uniswap v3 substreams SF built:
https://github.com/streamingfast/substreams-uniswap-v3 
Note that there is no need for be the subgraph schema and substreams proto definition to be the same: https://github.com/streamingfast/substreams-uniswap-v3/blob/develop/schema.graphql

Here is another (simpler) example you can refer to: https://github.com/azf20/scaling-adventure

this-username-is-taken:
In terms of "entities" and "fields", I prefer adding them iteratively over time, instead of defining everything at the start and leaving them empty, or having a super big PR that takes a long time to be merged in. This way it reduces mental burden and make things more manageable

this-username-is-taken:
Also, why start a new PR instead of continue off of https://github.com/messari/substreams/pull/101?

uniswap-v2/src/token_retriever.rs
```diff
@@ -0,0 +1,43 @@
...
+use substreams::Hex;
+
+use crate::abi::erc20;
+use crate::pb::dex_amm::v1::Token;
+
+pub(crate) struct TokenRetriever {
+    token_address: Vec<u8>,
+}
+
+impl TokenRetriever {
```
this-username-is-taken, line 11: This can be moved to the `substreams-helper` crate

uniswap-v2/src/token_retriever.rs
```diff
@@ -0,0 +1,43 @@
...
+            .unwrap_or(String::new())
+    }
+
+    pub(crate) fn get_decimals(&self) -> BigInt {
+        erc20::functions::Decimals {}
+            .call(self.token_address.to_vec())
+            .unwrap_or(BigInt::from(18))
+    }
+
+    pub(crate) fn to_struct(&self) -> Token {
```
this-username-is-taken, line 34: I wonder if it makes more sense to add this as a trait to `Token` (e.g. `from_address`) instead of creating a new struct. @robinbernon what do you think?

uniswap-v2/src/token_retriever.rs
```diff
@@ -0,0 +1,43 @@
...
+pub(crate) struct TokenRetriever {
+    token_address: Vec<u8>,
+}
+
+impl TokenRetriever {
+    pub(crate) fn new(token_address: Vec<u8>) -> Self {
+        TokenRetriever { token_address }
+    }
+
+    pub(crate) fn get_name(&self) -> String {
```
this-username-is-taken, line 16: does this need to be public?

uniswap-v2/src/pool_balance_retriever.rs
```diff
@@ -0,0 +1,51 @@
+use std::ops::Deref;
+
+use substreams::scalar::BigInt;
+use substreams::store::{DeltaBigInt, Deltas, StoreAdd, StoreAddBigInt};
+use substreams::Hex;
+use substreams_ethereum::NULL_ADDRESS;
+
+use crate::store_key::StoreKey;
+
+pub(crate) struct PoolBalanceRetriever<'a> {
```
this-username-is-taken, line 10: Why is this called `PoolBalanceRetriever`? Is it intended to be used to retrieve pool balance?

uniswap-v2/src/token_retriever.rs
robinbernon: Yeh I guess it depends what things you would want from the token address. If you want all this information here to create the Token struct every time then yes - something like `impl From<Vec<u8>> for Token` could work well here. If there are cases where you might want just some of the data from the address then I'd say it could be good to keep things as is.

common/proto/dex_amm.proto
harsh9200: yes, I have changed this to use already existing types.

uniswap-v2/src/common/traits.rs
```diff
@@ -0,0 +1,30 @@
+use ethabi::ethereum_types::Address;
+use substreams::store::{StoreGet, StoreGetProto};
+use substreams_helper::common::HasAddresser;
+use substreams_helper::hex::Hexable;
+
+use crate::pb::uniswap::v2::Pool;
+use crate::store_key::StoreKey;
+
+pub struct PoolAddresser<'a> {
+    pub store: &'a StoreGetProto<Pool>,
```
jaimehgb, line 10: HasAddresser is [already implemented for all stores](https://github.com/messari/substreams/blob/master/substreams-helper/src/common.rs#L35), including `StoreGetProto`, unless I did something wrong.

You can pass the store directly as a HasAddresser without implementing the trait again

uniswap-v2/Cargo.toml
```diff
@@ -17,6 +17,7 @@ pad = "0.1"
 substreams.workspace = true
 substreams-ethereum.workspace = true
 substreams-helper = { path = "../substreams-helper" }
+substreams-entity-change = "0.3.0"
```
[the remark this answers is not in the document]
harsh9200: I will update the version.

uniswap-v2/src/common/traits.rs
harsh9200: The key for the store is set to the address of the event. whereas I need to pass my own [key](https://github.com/messari/substreams/blob/39a889a296d35d2e0347446045778fe1e5e52631/uniswap-v2/src/common/traits.rs#L15-L17) to verify if the address is available in store or not.

uniswap-v2/Makefile
```diff
@@ -1,7 +1,16 @@
+.PHONY: build_all
+build_all:
```
robinbernon, line 2: erc20-holdings and maybe some other substreams in our repo have the same erc20-price dependency. Would you mind modifying their Makefiles to have the same cmds like this one also please?

uniswap-v2/src/abi.rs
```diff
@@ -1,15 +1,15 @@
 // DO NOT EDIT - the file is generated by build script
 #[rustfmt::skip]
 #[allow(unused_imports)]
-#[path = "../target/abi/erc20.rs"]
-pub mod erc20;
+#[path = "../target/abi/ERC20.rs"]
+pub mod ERC20;
```
robinbernon, line 5: We are getting warning here's for not following the standard syntax. Would it be possible to change this back to snakecase? (We can keep the filename uppercase if you like though?)

uniswap-v2/src/common/helpers.rs
```diff
@@ -0,0 +1,99 @@
...
+            name: self.get_name(),
+            symbol: self.get_symbol(),
+            address: self.0.to_hex(),
+            decimals: self.get_decimals().to_u64(),
+        }
+    }
+}
+
+impl Pool {
+    pub fn token0_ref(&self) -> Erc20Token {
```
robinbernon, line 46: ref?

uniswap-v2/src/common/traits.rs
```diff
@@ -0,0 +1,30 @@
...
+}
+
+impl<'a> PoolAddresser<'a> {
+    fn has_address(&self, key: Address) -> bool {
+        let pool = self
+            .store
+            .get_last(StoreKey::Pool.get_unique_pool_key(&key.to_hex()));
+
+        if pool.is_none() {
```
robinbernon, line 19: use `return pool.is_some()` instead of these 3 lines

uniswap-v2/src/modules/1_map_pool_created.rs
robinbernon, line 1: Mind if I ask a favour? Think it would be really nice if all the file get lists in numeric order. Would you mind changing all the single digit files to using two digits instead please? eg 1_map_pool_created.rs -> 01_map_pool_created.rs

uniswap-v2/src/modules/1_map_pool_created.rs
```diff
@@ -0,0 +1,48 @@
...
+}
+
+fn get_pools(block: &eth::Block, pools: &mut Vec<Pool>) {
+    let mut on_pair_created = |event: PairCreated, _tx: &eth::TransactionTrace, _log: &eth::Log| {
+        let pool = TokenContract::new(Address::from_slice(event.pair.as_slice())).as_struct();
+        let token0 = TokenContract::new(Address::from_slice(event.token0.as_slice())).as_struct();
+        let token1 = TokenContract::new(Address::from_slice(event.token1.as_slice())).as_struct();
+
+        pools.push(Pool {
+            name: token0.clone().symbol + "/" + token1.symbol.as_str(),
```
robinbernon, line 29: `token0.symbol.as_str()`

uniswap-v2/src/modules/1_map_pool_created.rs
```diff
@@ -0,0 +1,48 @@
...
+            symbol: String::new(),
+            address: pool.clone().address,
```
robinbernon, line 31: `pool.address.clone()`

uniswap-v2/src/modules/4_map_pool_events.rs
```diff
@@ -0,0 +1,175 @@
...
+
+#[substreams::handlers::map]
+pub fn map_pool_events(
+    block: eth::Block,
+    pools_store: StoreGetProto<Pool>,
+    balance_deltas: Deltas<DeltaBigInt>,
+) -> Result<Events, substreams::errors::Error> {
+    let mut events = vec![];
+
+    handle_mint(&block, &pools_store, &balance_deltas, &mut events);
```
robinbernon, line 24: Check here: https://github.com/messari/substreams/blob/master/substreams-helper/src/event_handler.rs#L13-L19

Can you change these fns to just set up the handlers instead for the different events and then call `eh.handle_events();` after. This way you will loop through all the logs just once

uniswap-v2/proto/v1/uniswap.proto
```diff
@@ -0,0 +1,65 @@
...
+  optional string output_token_amount = 2;
+}
+
+message WithdrawEvent {
+  repeated string input_token_amounts = 1;
+  optional string output_token_amount = 2;
+}
+
+message SyncEvent {
+  string reserve0 = 1;
```
robinbernon, line 55: @jaimehgb Didn't Alex say that the substreams CLI now has support for debugging BigInts? Shall we switch back to BigInt types now?

uniswap-v2/src/modules/7_store_usd_prices.rs
```diff
@@ -0,0 +1,170 @@
...
+) -> BigDecimal {
+    if token_address.eq(constants::WETH_ADDRESS) {
+        return get_weth_price_in_usd(ordinal, native_prices_store);
+    }
+    if constants::STABLE_COINS.contains(&token_address.as_str()) {
+        return BigDecimal::one();
+    }
+
+    for address in constants::PAIR_COINS.into_iter() {
+        let whitelisted_token = address.to_string();
```
robinbernon, line 106: whitelisted_token is only used by reference from here. Could you just use `let whitelisted_token = address;` instead?

uniswap-v2/src/modules/7_store_usd_prices.rs
```diff
@@ -0,0 +1,170 @@
...
+            ordinal,
+            StoreKey::TokenBalance.get_unique_pair_key(address1, address2),
+        )
+        .unwrap_or(BigDecimal::zero());
+
+    if address2.eq(constants::WETH_ADDRESS) {
+        amount_locked_usd = amount_locked_usd * get_weth_price_in_usd(ordinal, native_prices_store);
+    }
+
+    if amount_locked_usd.ge(&BigDecimal::from_str("5000").unwrap()) {
```
robinbernon, line 153: No if required here. Just use `amount_locked_usd.ge(&BigDecimal::from_str("5000").unwrap())` instead

uniswap-v2/src/modules/7_store_usd_prices.rs
```diff
@@ -0,0 +1,170 @@
...
+
+    return false;
+}
+
+fn get_weth_price_in_usd(ordinal: u64, native_prices_store: &StoreGetBigDecimal) -> BigDecimal {
+    native_prices_store
+        .get_at(
+            ordinal,
+            StoreKey::TokenPrice.get_unique_pair_key(
+                &constants::WETH_ADDRESS.to_string(),
```
robinbernon, line 165: Use `constants::WETH_ADDRESS` instead of `&constants::WETH_ADDRESS.to_string()`

uniswap-v2/src/modules/9_store_volume.rs
```diff
@@ -0,0 +1,97 @@
...
+    prices_store: StoreGetBigDecimal,
+    output_store: StoreAddBigDecimal,
+) {
+    for event in swap_events.events {
+        match event.r#type.unwrap() {
+            SwapType(swap_event) => {
+                let pool_address = event.pool;
+                let ordinal = event.log_ordinal;
+
+                let token_in = swap_event.token_in.unwrap();
```
robinbernon, line 25: `swap_event.token_in.as_ref().unwrap()` will avoid taking value from `swap_event`

uniswap-v2/src/modules/9_store_volume.rs
```diff
@@ -0,0 +1,97 @@
...
+                    &swap_event.amount_in,
+                    &swap_event.amount_out,
+                    &prices_store,
+                );
+
+                output_store.add(
+                    ordinal,
+                    StoreKey::VolumeByTokenUSD
+                        .get_unique_pair_key(&pool_address, &token_in.address),
+                    &volume.clone(),
```
robinbernon, line 41: Remove `clone()` here and same for other cases below

uniswap-v2/src/modules/9_store_volume.rs
```diff
@@ -0,0 +1,97 @@
...
+    token_in: &Erc20Token,
+    token_out: &Erc20Token,
+    amount_in: &String,
+    amount_out: &String,
+    prices_store: &StoreGetBigDecimal,
+) -> BigDecimal {
+    let token_in_price = utils::get_token_price(ordinal, prices_store, &token_in.address);
+    let token_out_price = utils::get_token_price(ordinal, prices_store, &token_out.address);
+
+    let amount_in = BigInt::try_from(amount_in)
```
robinbernon, line 72: Is there a `.from(_)` instead to avoid the unwrap?

uniswap-v2/src/utils.rs
```diff
@@ -0,0 +1,59 @@
...
+    let new_value = delta.new_value.clone();
+
+    return new_value.clone().sub(old_value);
+}
+
+pub fn calculate_revenue(volume: BigDecimal) -> (BigDecimal, BigDecimal) {
+    let supply_side_revenue =
+        volume.clone() * BigDecimal::from(25_i32) / BigDecimal::from(10000_i32);
+    let protocol_side_revenue =
+        volume.clone() * BigDecimal::from(5_i32) / BigDecimal::from(10000_i32);
```
robinbernon, line 35: Can remove this clone

uniswap-v2/src/utils.rs
```diff
@@ -0,0 +1,59 @@
...
+pub fn is_pricing_asset(address: &String) -> bool {
+    constants::STABLE_COINS.contains(&address.as_str())
+        || constants::WHITELIST_TOKENS.contains(&address.as_str())
+}
+
+pub fn delta_value(delta: &DeltaBigDecimal) -> BigDecimal {
+    let old_value = delta.old_value.clone();
+    let new_value = delta.new_value.clone();
+
+    return new_value.clone().sub(old_value);
```
robinbernon, line 28: Double clone

uniswap-v2/src/modules/10_store_volume_by_token_amount.rs
```diff
@@ -0,0 +1,67 @@
...
+                let ordinal = event.log_ordinal;
+                let pool_address = event.pool;
+
+                let day_id = utils::get_day_id(event.timestamp as i64);
+                let hour_id = utils::get_hour_id(event.timestamp as i64);
+
+                let token_in = swap.token_in.unwrap();
+                let token_out = swap.token_out.unwrap();
+
+                let amount_in = BigInt::try_from(swap.amount_in).unwrap();
```
robinbernon, line 23: from()?

uniswap-v2/src/modules/11_store_cumulative_fields.rs
```diff
@@ -0,0 +1,40 @@
...
+    volume_deltas: Deltas<DeltaBigDecimal>,
+    output_store: StoreAddBigDecimal,
+) {
+    for delta in volume_deltas.deltas {
+        if let Some(pool_address) = StoreKey::Volume.get_pool(&delta.key) {
+            let volume = utils::delta_value(&delta);
+            let (supply_side_revenue, protocol_side_revenue) =
+                utils::calculate_revenue(volume.clone());
+
+            output_store.add(
```
robinbernon, line 18: Putting this line before `let (supply_side_revenue, protocol_side_revenue) = utils::calculate_revenue(volume.clone());` will allow you to avoid cloning volume variable

uniswap-v2/src/modules/11_store_cumulative_fields.rs
```diff
@@ -0,0 +1,40 @@
...
+                delta.ordinal,
+                StoreKey::CumulativeVolumeUSD.get_unique_pool_key(&pool_address),
+                &volume,
+            );
+            output_store.add(
+                delta.ordinal,
+                StoreKey::CumulativeSupplySideRevenueUSD.get_unique_pool_key(&pool_address),
+                &supply_side_revenue.clone(),
```
robinbernon, line 26: Remove clone

uniswap-v2/src/modules/11_store_cumulative_fields.rs
```diff
@@ -0,0 +1,40 @@
...
+            );
+            output_store.add(
+                delta.ordinal,
+                StoreKey::CumulativeProtocolSideRevenueUSD.get_unique_pool_key(&pool_address),
+                &protocol_side_revenue.clone(),
```
robinbernon, line 31: Same here

uniswap-v2/src/modules/12_map_liquidity_pool_entity.rs
```diff
@@ -0,0 +1,160 @@
...
+    balances_store: StoreGetBigInt,
+    balances_deltas: Deltas<DeltaBigInt>,
+    pool_tvl_store: StoreGetBigDecimal,
+    cumulative_fields_store: StoreGetBigDecimal,
+) -> Result<EntityChanges, ()> {
+    let mut entity_changes: Vec<EntityChange> = vec![];
+
+    for delta in balances_deltas.deltas.iter() {
+        if let Some(pool_address) = StoreKey::TotalBalance.get_pool(&delta.key) {
+            let mut is_initialized: bool = true;
```
robinbernon, line 25: `let is_initialized = delta.operation != store_delta::Operation::Create;`

uniswap-v2/src/store_key.rs
```diff
@@ -0,0 +1,144 @@
...
+
+    pub fn get_unique_pair_key(&self, key1: &String, key2: &String) -> String {
+        format!("{}:{}:{}", self.unique_id(), key1, key2)
+    }
+
+    pub fn get_unique_protocol_key(&self) -> String {
+        format!("[Protocol]:{}", self.unique_id())
+    }
+
+    pub fn get_unique_daily_pool_key(&self, day_id: BigInt, key: &String) -> String {
```
robinbernon, line 48: Accept `&BigInt` arg instead and then all uses of fn should take refs instead. Same for other fn below

uniswap-v2/src/modules/13_store_daily_and_hourly_fields.rs
```diff
@@ -0,0 +1,99 @@
...
+        let volume = utils::delta_value(&delta);
+
+        if let Some(pool_address) = StoreKey::Volume.get_pool(&delta.key) {
+            let (supply_side_revenue, protocol_side_revenue) =
+                utils::calculate_revenue(volume.clone());
+
+            output_store.add(
+                ordinal,
+                StoreKey::HourlyVolumeUSD
+                    .get_unique_hourly_pool_key(hour_id.clone(), &pool_address),
```
robinbernon, line 30: Have already added comment about `StoreKey::HourlyVolumeUSD.get_unique_hourly_pool_key` so make sure to remove all these clones once fn is modded and change these to refs instead :)

uniswap-v2/src/modules/14_map_liquidity_pool_snapshots_entity.rs
```diff
@@ -0,0 +1,500 @@
...
+    balances_store: &StoreGetBigInt,
+    pool_tvl_store: &StoreGetBigDecimal,
+    cumulative_fields_store: &StoreGetBigDecimal,
+    daily_and_hourly_fields_store: &StoreGetBigDecimal,
+    volume_by_token_amount_store: &StoreGetBigInt,
+    prices_store: &StoreGetBigDecimal,
+    block_number: &BigInt,
+    timestamp: &BigInt,
+) -> EntityChange {
+    let id = [pool.clone().address, day_id.clone().to_string()].join("-");
```
robinbernon, line 94: Haven't checked to see if it works but try like: `[pool.address.as_str(), day_id.to_string().as_str()].join("-");` instead

uniswap-v2/src/modules/14_map_liquidity_pool_snapshots_entity.rs
```diff
@@ -0,0 +1,500 @@
...
+    pool_supply_store: &StoreGetBigInt,
+    balances_store: &StoreGetBigInt,
+    pool_tvl_store: &StoreGetBigDecimal,
+    cumulative_fields_store: &StoreGetBigDecimal,
+    daily_and_hourly_fields_store: &StoreGetBigDecimal,
+    volume_by_token_amount_store: &StoreGetBigInt,
+    block_number: &BigInt,
+    timestamp: &BigInt,
+) -> EntityChange {
+    let id = [pool.clone().address, hour_id.clone().to_string()].join("-");
```
robinbernon, line 314: Same as above

uniswap-v2/src/modules/18_map_protocol_entity.rs
```diff
@@ -0,0 +1,115 @@
...
+pub fn map_protocol_entity(
+    input_token_balances_deltas: Deltas<DeltaBigInt>,
+    protocol_tvl_store: StoreGetBigDecimal,
+    protocol_cumulative_fields_store: StoreGetBigDecimal,
+) -> Result<EntityChanges, ()> {
+    let mut entity_changes: Vec<EntityChange> = vec![];
+
+    for delta in input_token_balances_deltas.deltas.iter() {
+        if let Some(_) = StoreKey::TotalBalance.get_pool(&delta.key) {
+            let mut is_initialized: bool = true;
```
robinbernon, line 20: Same thing as I mentioned previously

uniswap-v2/src/modules/19_map_financial_daily_snapshot_entity.rs
```diff
@@ -0,0 +1,167 @@
...
+    day_id: BigInt,
+    protocol_tvl_store: &StoreGetBigDecimal,
+    protocol_cumulative_fields_store: &StoreGetBigDecimal,
+    protocol_daily_fields_store: &StoreGetBigDecimal,
+    block_number: &BigInt,
+    timestamp: &BigInt,
+) -> EntityChange {
+    let id = [
+        constants::UNISWAP_V2_FACTORY.to_string(),
+        day_id.clone().to_string(),
```
robinbernon, line 66: Remove clone

uniswap-v2/src/modules/21_map_events_entity.rs
```diff
@@ -0,0 +1,206 @@
...
+#[substreams::handlers::map]
+pub fn map_events_entity(
+    pool_events_map: Events,
+    pool_store: StoreGetProto<Pool>,
+    prices_store: StoreGetBigDecimal,
+) -> Result<EntityChanges, ()> {
+    let mut entity_changes: Vec<EntityChange> = vec![];
+
+    for event in pool_events_map.events {
+        let ordinal = event.clone().log_ordinal as u64;
```
robinbernon, line 24: No clone needed

uniswap-v2/src/modules/21_map_events_entity.rs
```diff
@@ -0,0 +1,206 @@
...
+            )),
+            _ => {}
+        }
+    }
+
+    Ok(EntityChanges { entity_changes })
+}
+
+fn get_event_id(event: &Event) -> String {
+    [event.hash.clone(), event.log_index.to_string()].join("-")
```
robinbernon, line 58: Same thing here as previous cases

uniswap-v2/src/modules/21_map_events_entity.rs
```diff
@@ -0,0 +1,206 @@
...
+fn calculate_event_amount_usd(
+    ordinal: u64,
+    input_tokens: Vec<Erc20Token>,
+    amounts: Vec<String>,
+    store: &StoreGetBigDecimal,
+) -> BigDecimal {
+    let mut amount_usd = BigDecimal::zero();
+
+    for (idx, token) in input_tokens.iter().enumerate() {
+        let token_amount = BigInt::try_from(amounts[idx].clone()).unwrap();
```
robinbernon, line 70: from()

uniswap-v2/src/modules/21_map_events_entity.rs
```diff
@@ -0,0 +1,206 @@
...
+    event: &Event,
+    deposit: &DepositEvent,
+    prices_store: &StoreGetBigDecimal,
+) -> EntityChange {
+    let id = get_event_id(event);
+
+    let mut deposit_entity_change =
+        EntityChange::new("Deposit", id.as_str(), ordinal, Operation::Create);
+
+    let input_tokens = pool.input_tokens.clone().unwrap().items;
```
robinbernon, line 91: `let input_tokens = pool.input_tokens.as_ref().unwrap().items.clone();`

uniswap-v2/src/modules/21_map_events_entity.rs
```diff
@@ -0,0 +1,206 @@
...
+    let input_token_amounts = deposit.input_token_amounts.clone();
+    let output_token_amount =
+        BigInt::try_from(deposit.output_token_amount.clone().unwrap()).unwrap();
```
robinbernon, line 94: Same thing as above

uniswap-v2/src/modules/21_map_events_entity.rs
```diff
@@ -0,0 +1,206 @@
...
+    deposit: &DepositEvent,
+    prices_store: &StoreGetBigDecimal,
+) -> EntityChange {
+    let id = get_event_id(event);
+
+    let mut deposit_entity_change =
+        EntityChange::new("Deposit", id.as_str(), ordinal, Operation::Create);
+
+    let input_tokens = pool.input_tokens.clone().unwrap().items;
+    let input_token_amounts = deposit.input_token_amounts.clone();
```
robinbernon, line 92: Remove clone

uniswap-v2/src/modules/21_map_events_entity.rs
```diff
@@ -0,0 +1,206 @@
...
+    let amount_usd = calculate_event_amount_usd(
+        ordinal,
+        input_tokens,
+        input_token_amounts.clone(),
+        prices_store,
+    );
+
+    deposit_entity_change
+        .change("id", id)
+        .change("hash", event.clone().hash)
```
robinbernon, line 105: Clone field rather than entire event - do for other cases below also

uniswap-v2/src/modules/21_map_events_entity.rs
```diff
@@ -0,0 +1,206 @@
...
+) -> EntityChange {
+    let id = get_event_id(event);
+
+    let mut withdraw_entity_change: EntityChange =
+        EntityChange::new("Withdraw", id.as_str(), ordinal, Operation::Create);
+
+    let input_tokens = pool.input_tokens.clone().unwrap().items;
+    let input_token_amounts = withdraw.input_token_amounts.clone();
+    let output_token_amount =
+        BigInt::try_from(withdraw.output_token_amount.clone().unwrap()).unwrap();
```
robinbernon, line 137: `withdraw.output_token_amount.as_ref().unwrap().clone()` and maybe also a from here too?

uniswap-v2/src/modules/21_map_events_entity.rs
```diff
@@ -0,0 +1,206 @@
...
+    withdraw: &WithdrawEvent,
+    prices_store: &StoreGetBigDecimal,
+) -> EntityChange {
+    let id = get_event_id(event);
+
+    let mut withdraw_entity_change: EntityChange =
+        EntityChange::new("Withdraw", id.as_str(), ordinal, Operation::Create);
+
+    let input_tokens = pool.input_tokens.clone().unwrap().items;
+    let input_token_amounts = withdraw.input_token_amounts.clone();
```
robinbernon, line 135: This and above can get changed like in fn above

uniswap-v2/src/modules/21_map_events_entity.rs
```diff
@@ -0,0 +1,206 @@
...
+        input_token_amounts.clone(),
+        prices_store,
+    );
+
+    withdraw_entity_change
+        .change("id", id)
+        .change("hash", event.clone().hash)
+        .change("logIndex", event.log_index as i32)
+        .change("protocol", constants::UNISWAP_V2_FACTORY.to_string())
+        .change("to", event.clone().to)
```
robinbernon, line 151: Same stuff here also

uniswap-v2/src/modules/21_map_events_entity.rs
```diff
@@ -0,0 +1,206 @@
...
+    event: &Event,
+    swap: &SwapEvent,
+    prices_store: &StoreGetBigDecimal,
+) -> EntityChange {
+    let id = get_event_id(event);
+
+    let mut swap_entity_change: EntityChange =
+        EntityChange::new("Swap", id.as_str(), ordinal, Operation::Create);
+
+    let token_in = swap.clone().token_in.unwrap();
```
robinbernon, line 176: clone order

uniswap-v2/substreams.yaml
```diff
@@ -1,101 +1,241 @@
...
+    valueType: bigdecimal
+    inputs:
+      - store: store_pools
+      - map: map_pool_events
+      - store: store_native_prices
+
+  - name: store_pool_tvl
+    kind: store
+    updatePolicy: set
+    valueType: bigdecimal
```
robinbernon, line 88: So there are a few cases of modules that are like this one. This module for example could be combined with the module above as they are both performing the same store operation without any temporal based conflicts. If they got run together as one module you would only need 1 store squashing event rather than 2 so would save a fair bit of processing if it could be done this way. Two ways to go about this:

1. Change the above module name to be inclusive of this module as well with a name like store_usd_prices_and_pool_tvl and put the fns in the same file.

2. Similar to above but keep store_pool_tvl.rs file in modules folder and don't derive the store macro for it. Instead pass it the store and relevant args from the store_usd_prices module and we can "pretend" it is it's own module even though it's actually just an fn now. (Should help keep each module relatively "clean" even when a lot is going on)

uniswap-v2/src/modules/10_store_volume_by_token_amount.rs
harsh9200: `From<String>` is not implemented by bigint, that's why I had to use `try_from` here.

uniswap-v2/src/modules/9_store_volume.rs
harsh9200: No, the from<String> is not implemented for bigint

uniswap-v2/src/abi.rs
harsh9200: Can we keep this as it is? We can probably add `#[allow(non_snake_case)]` to allow non snake case names in codegen.rs

uniswap-v2/src/modules/1_map_pool_created.rs
harsh9200: Sure

uniswap-v2/src/abi.rs
robinbernon: Ok sure we can do that then

uniswap-v2/src/modules/7_store_usd_prices.rs
robinbernon: Can you change the signature for it? (All other calls will be unaffected by the change as there is auto coersion from &String to &str

uniswap-v2/src/modules/4_map_pool_events.rs
harsh9200: As I have to mutably store all the events in a list, I have to do it in separate functions. 

Ref: https://github.com/messari/substreams/blob/master/substreams-helper/src/event_handler.rs#L22-L39

uniswap-v2/src/modules/7_store_usd_prices.rs
harsh9200: Ah, I see, that works. Thanks

Issue #94, opened 2022-10-25
substream to calculate eth total supply


this-username-is-taken:
apparently this is not trivial. some reference:

https://github.com/lastmjs/eth-total-supply
https://twitter.com/mhswende/status/1292370934658142211
https://decrypt.co/38271/so-what-is-the-ethereum-eth-total-supply

PR #118, opened 2022-12-22
feat(#major); Token Standard

dmelotik:
Background: We want to get historical token balances on multiple blockchains over multiple tokens (erc20's, native tokens, etc.) In order to streamline our ingestion of this data we want to standardize the proto definitions for this.

Deliverables:
- Standard `token` proto definition
- Build the `eth-balance` substream so that it uses this proto standard
- Define the standard in documentation for others to use


common/README.md
```diff
@@ -0,0 +1,69 @@
...
+
+## Token Proto
+
+The token proto is designed to store the token balances of all accounts throughout the blockchains history. It is designed to be network-agnostic and support all tokens. Token's can be blockchain native, erc20s, or any other derivative.
+
+### Definitions
+
+The first part are the token definitions. This is simple, and defines the tokens that are supported in the substream implementation.
+
+> Note: if the token is a native token (e.g. ethereum), then the address is the `0x0` address and the other values need to be hardcoded.
```
this-username-is-taken, line 21: IIRC, bytes by default serializes without the `0x` prefix in Rust. They are kept this way when we use them in substreams to avoid additional parsing/padding. Any particular reason to store them with `0x`?

common/README.md
```diff
@@ -0,0 +1,69 @@
...
+
+TODO
+
+## DEX AMM Proto
+
+TODO
+
+## Token Proto
+
+The token proto is designed to store the token balances of all accounts throughout the blockchains history. It is designed to be network-agnostic and support all tokens. Token's can be blockchain native, erc20s, or any other derivative.
```
this-username-is-taken, line 15: What about NFTs? Would they just adopt the `Token` definition as well or would they have a separate one?

common/README.md
dmelotik: Will look into this tomo

common/README.md
dmelotik: I should fix the verbiage, this was meant to be only for fungible tokens. And it could fit the erc20, maybe not, and to be inclusive to other blockchain standards like solana SPL tokens.

I think it is best to make a separate proto def for NFTs, I think from our experience it is best to make schemas focused. That approach also feels much more native to substreams. Would you agree?

common/README.md
dmelotik: Removed that since it is not something I tackled in this PR

common/README.md
dmelotik: lmk what you think of the proto named evm.token.proto? Do you think it should like in that folder or should I make the path `./common/proto/token/evm.token.proto` assuming we will have more

Cargo.toml
```diff
@@ -19,9 +19,6 @@ substreams = "0.3.2"
 substreams-ethereum = "0.6.2"
 substreams-solana = "0.1.0"
 
-[build]
-target = "wasm32-unknown-unknown"
```
this-username-is-taken, line 23: why remove this?

eth-balance/src/lib.rs
```diff
@@ -21,13 +24,34 @@ fn store_balance(block: pbeth::v2::Block, output: store::StoreSetRaw) {
...
     }
+
+    Ok(token::Accounts { items: accounts })
+}
+
+//////////////////////////
+//// Helper Functions ////
+//////////////////////////
+
+fn get_eth_token() -> token::Token {
```
this-username-is-taken, line 47: should move this to `substreams-helper`

common/proto/evm.token.proto
```diff
@@ -0,0 +1,45 @@
...
+message Transfer {
+  string tx_hash = 1;
+  optional uint32 log_index = 2;
+  optional uint64 log_ordinal = 3;
+  Token token = 4;
+  string from = 5;
+  string to = 6;
+  string amount = 7; // BigInt, in token's native amount
+  optional string amount_usd = 8;
+  optional string reason = 9;
```
this-username-is-taken, line 30: is this needed?

common/proto/evm.token.proto
```diff
@@ -0,0 +1,45 @@
...
+  optional string total_supply = 5;
+}
+
+message Transfers {
+  repeated Transfer items = 1;
+}
+
+message Transfer {
+  string tx_hash = 1;
+  optional uint32 log_index = 2;
```
this-username-is-taken, line 23: why are `log_index` and `log_ordinal` optional? what are some scenarios these will be missing?

common/proto/evm.token.proto
```diff
@@ -0,0 +1,45 @@
...
+message Tokens {
+  repeated Token items = 1;
+}
+
+message Token {
+  string address = 1;
+  string name = 2;
+  string symbol = 3;
+  uint64 decimals = 4;
+  optional string total_supply = 5;
```
this-username-is-taken, line 14: I would separate `total_supply` from `Token` as all other fields are immutable. One way to do it is to represent it using `TokenBalance` instead.

common/proto/evm.token.proto
```diff
@@ -0,0 +1,45 @@
...
+  optional uint64 log_ordinal = 3;
+  Token token = 4;
+  string from = 5;
+  string to = 6;
+  string amount = 7; // BigInt, in token's native amount
+  optional string amount_usd = 8;
+  optional string reason = 9;
+}
+
+message TokenBalance {
```
this-username-is-taken, line 33: I'm not sure what's your conceptual representation of a `TokenBalance`. Maybe it'll be helpful to document it. It seems to me that this is a spot balance which is specific to a particular block/timestamp. If that's the case I'd include block height and timestamp in the message as well

Cargo.toml
dmelotik: for some reason I couldn't build with it in there. But I fixed it

common/proto/evm.token.proto
dmelotik: it was based off this: https://github.com/streamingfast/firehose-ethereum/blob/develop/proto/sf/ethereum/type/v2/type.proto#L411 would really only be used in eth-balance

common/proto/evm.token.proto
dmelotik: I made them optional when I was going to make one proto for all chains, Updating that

common/proto/evm.token.proto
dmelotik: included

eth-balance/src/lib.rs
dmelotik: Moved, but still struggling to match the type

common/proto/evm.token.proto
this-username-is-taken: I'm not sure how interesting it is to include this. For most of our use cases so far, I don't think we've needed reason much if at all. Although we also never had access to this piece of this information. @yulesa what are your thoughts on this? Do you think there will be places where this piece of information can be useful?

common/proto/evm.token.proto
this-username-is-taken: exactly

common/proto/evm.token.proto
yulesa: I have no idea what this is. I need context.

dmelotik: I talked to @yulesa and "yes", but I also need to rethink how we want to show the eth balances, because there is more nuance.

ie: instead of thinking about it in the sense of balances at x block we should think about it in terms of balances and changes tied to calls

common/proto/evm.token.proto
dmelotik: Since total_supply can change do you want a block number associated with it? is that the thing?

The way I restructured the proto, it doesn't feel right to put it in TokenBalance

dmelotik: If the block associated with the total supply is the issue, wouldn't we be able to get the block number on the block that  outputted that message in the output?

eth-balance/src/lib.rs
```diff
@@ -1,33 +1,50 @@
...
-                output.set(
-                    transaction.end_ordinal,
-                    format!("Address:{}", Hex(&balance_change.address).to_string()),
-                    &new_value.to_string(),
-                )
+                balance_changes.push(token::TokenBalance {
+                    log_ordinal: balance_change.ordinal,
+                    token: token_helper::get_eth_token(),
+                    address: Hex(&balance_change.address).to_string(),
+                    old_balance: token_helper::bigint_to_string(balance_change.old_value.clone()),
```
robinbernon, line 22: Impl Display trait for bigint and use instead. We can have the impl in substreams helper also

eth-balance/src/lib.rs
```diff
@@ -1,33 +1,50 @@
...
+                    new_balance: token_helper::bigint_to_string(balance_change.new_value.clone()),
+                    reason: Some(balance_change.reason),
+                });
             }
         }
+        transfers.push(token::Transfer {
+            tx_hash: Hex(&transaction.hash).to_string(),
```
robinbernon, line 29: hex::encode()

eth-balance/substreams.yaml
```diff
@@ -8,19 +8,20 @@ imports:
...
     file: "../target/wasm32-unknown-unknown/release/eth_balance.wasm"
 
 modules:
-  - name: store_balance
-    kind: store
-    updatePolicy: set
-    valueType: string
+  - name: map_balances
+    kind: map
+    initialBlock: 0
```
robinbernon, line 23: You can remove this

substreams-helper/build.rs
```diff
@@ -2,6 +2,8 @@ use anyhow::{Ok, Result};
 use substreams_common::codegen;
 
 fn main() -> Result<(), anyhow::Error> {
+    codegen::generate(None)?;
     codegen::generate_abi(None)?;
```
robinbernon, line 6: Remove this

substreams-helper/src/token.rs
```diff
@@ -0,0 +1,25 @@
...
+use num_bigint;
+use substreams::scalar::BigInt;
+use substreams_ethereum::pb::eth as pbeth;
+
+pub fn get_eth_token() -> Option<Token> {
+    let eth_token = Token {
+        address: "0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE".to_string(),
+        name: "Ethereum".to_string(),
+        symbol: "ETH".to_string(),
+        decimals: 18 as u64,
```
robinbernon, line 11: Avoid type conversion - probs fine as compiler will optimise away although better to use 18_u64 instead.

substreams-helper/src/token.rs
```diff
@@ -0,0 +1,25 @@
...
+        total_supply: None,
+    };
+
+    Some(eth_token)
+}
+
+// TODO: replace this with substreams::scalar::BigInt once the wrapper is integrated
+pub fn bigint_to_string(number: Option<pbeth::v2::BigInt>) -> String {
```
robinbernon, line 19: As mentioned above this can be changed to an impl of Display instead

eth-balance/src/lib.rs
```diff
@@ -1,33 +1,50 @@
 #[rustfmt::skip]
 pub mod pb;
 
-use num_bigint;
-use substreams::scalar::BigInt;
-use substreams::store::{StoreNew, StoreSet, StoreSetRaw};
-use substreams::{store, Hex};
+use substreams_helper::pb::evm_token::v1 as token;
+// use pb::evm_token::v1 as token;
```
robinbernon, line 5: I guess you can remove this now also :)

eth-balance/src/pb/messari.token.v1.rs
```diff
@@ -0,0 +1,217 @@
+// @generated
+#[derive(Clone, PartialEq, ::prost::Message)]
+pub struct Tokens {
```
robinbernon, line 3: This file should not be in the PR - Use build.rs to push this to target folder like the other substreams projects do

common/README.md
this-username-is-taken: Why `evm.token` instead of `erc20`? ERC20 is more of a convention and I haven't seen anyone used the term EVM token. This can cause confusion.

common/proto/evm.token.proto
this-username-is-taken: I'd imagine `Token` (without total supply) is a very commonly needed/used data structure across many substreams, whereas total supply is used much less. If you include total supply in `Token`, every change to total supply will lead to a new message sent over to the downstream modules, which will lead to lots of unnecessary data and compute incurred to the downstream modules.

common/README.md
dmelotik: Yeah I guess only ETH, I will change to erc20

common/proto/evm.token.proto
dmelotik: For now I am removing it from the eth-balance proto, I think if we want to do that we should have a separate substream to calculate the total supply of eth. We could consider adding it into the erc20 substream if that is important.

substreams-helper/build.rs
dmelotik: How are the erc20 abis retrieved then? The erc20 abi is needed for the erc20 functions in `substreams-helper`

substreams-helper/src/token.rs
dmelotik: Looks like it works, but what is this `_u64` part? I can't find anything in the rust docs about using the `_`

common/proto/evm.token.proto
this-username-is-taken: yeah, having a separate one sounds good to me. I actually had an issue open for this earlier: https://github.com/messari/substreams/issues/94

PR #120, opened 2023-01-11
fix(); erc20 holding; update config

dmelotik:
Update the config to be the same as other substreams

- Adding `rust-toolchain.toml` is not necessary, but we have it in all the other substreams.
- I was running into issues because I didn't have the wasm target installed
- @robinbernon do you know why that uniswap section is being added when I build erc20-holdings? Should it be there now?

### DIFF
--- a/erc20-holdings/src/lib.rs
+++ b/erc20-holdings/src/lib.rs
@@ -46,6 +46,7 @@ fn map_block_to_erc20_contracts(
         if call.call_type == pbeth::v2::CallType::Create as i32 {
             // skipping contracts that are too short to be an erc20 token
             if code_len(call) < 150 {
+                // contract byte code < 150 bytes
                 continue;
             }
 
--- a/erc20-holdings/src/lib.rs
+++ b/erc20-holdings/src/lib.rs
@@ -25,7 +25,7 @@ use substreams_ethereum::{pb::eth as pbeth, Event, NULL_ADDRESS};
 use substreams_helper::keyer::chainlink_asset_key;
 use substreams_helper::types::Address;
 
-fn code_len(call: &pbeth::v2::Call) -> usize {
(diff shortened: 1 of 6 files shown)


erc20-holdings/substreams.yaml
```diff
@@ -22,7 +22,7 @@ binaries:
 modules:
   - name: map_block_to_erc20_contracts
     kind: map
-    initialBlock: 10606500
+    initialBlock: 1
```
robinbernon, line 25: Slight misalignment with this initial block and the make run cmd (cmd starts at 0 rather than 1) - perhaps you can change one or the other to make both line up?

erc20-holdings/src/lib.rs
```diff
@@ -46,6 +46,7 @@ fn map_block_to_erc20_contracts(
         if call.call_type == pbeth::v2::CallType::Create as i32 {
             // skipping contracts that are too short to be an erc20 token
             if code_len(call) < 150 {
+                // contract byte code < 150 bytes
```
robinbernon, line 49: Not a fan of comment as it's too similar looking to the above. Think maybe it could be better to change the fn name from code_len to something like contract_bytecode_len() instead?

dmelotik: agreed, fixed

erc20-holdings/substreams.yaml
dmelotik: I updated the cmd to start at 1, before it didn't have any start block so it just started at 1 by default

this-username-is-taken (approved):
Next thing to do is to think of a test/validation strategy for erc20 tokens, but that can be done in a separate PR

dmelotik:
@this-username-is-taken I have this document here: https://www.notion.so/messari/ERC-20-Contracts-6fcac1983b99450b97670817a27a3159

PR #124, opened 2023-01-19
feat(); SPL Tokens; new substream

dmelotik:
New substream to get SPL token balances on solana.

Deliverables:
- Substream module following `sol_token.proto` to get the balances of all spl tokens on solana

Model off: https://github.com/streamingfast/substreams-playground/tree/master/modules/sol-spl-tokens

### DIFF
--- a/spl-balance/README.md
+++ b/spl-balance/README.md
@@ -1,0 +1,3 @@
+# Solana SPL Token Balances
+
+This substream is designed to get the token balances of all SPL tokens on Solana.


solana-sample/src/abi.rs
```diff
@@ -0,0 +1 @@
+// DO NOT EDIT - the file is generated by build script
```
robinbernon, line 1: rm

solana-sample/rust-toolchain.toml
```diff
@@ -0,0 +1,4 @@
+[toolchain]
```
robinbernon, line 1: Not really a fan of local toolchains for each project. Would be best to have these settings configured globally for all projects by default if possible

solana-sample/src/pb.rs
```diff
@@ -0,0 +1,19 @@
+#[rustfmt::skip]
+#[path = "../target/pb/messari.common.v1.rs"]
+pub(in crate::pb) mod common_v1;
```
robinbernon, line 3: proto files aren't being used so we could remove them from the substream for now?

solana-sample/src/lib.rs
```diff
@@ -1,7 +1,8 @@
-use substreams::{log, store};
+use substreams::pb;
```
robinbernon, line 1: Same deal here

solana-sample/substreams.yaml
```diff
@@ -6,15 +6,22 @@ package:
 imports:
   sol: https://github.com/streamingfast/firehose-solana/releases/download/v0.1.0/solana-v0.1.0.spkg
 
+protobuf:
+  files:
+    - test.proto
```
robinbernon, line 11: rm?

spl-balance/src/abi.rs
```diff
@@ -0,0 +1 @@
+// DO NOT EDIT - the file is generated by build script
```
robinbernon, line 1: rm

spl-balance/rust-toolchain.toml
```diff
@@ -0,0 +1,4 @@
+[toolchain]
```
robinbernon, line 1: rm if possible

spl-balance/src/instruction.rs
```diff
@@ -0,0 +1,658 @@
+use crate::option::COption;
+use std::convert::TryInto;
+use substreams::errors::Error;
+
+/// Instructions supported by the token program.
+#[repr(C)]
+#[derive(Clone, Debug, PartialEq)]
+pub enum TokenInstruction<'a> {
```
robinbernon, line 8: Are you planning to use logic from this file later on?

spl-balance/src/lib.rs
```diff
@@ -0,0 +1,178 @@
+pub mod instruction;
+pub mod pb;
+
+use bs58;
+use pb::sol_token::v1 as proto;
+use substreams::log;
+use substreams::store::{StoreGet, StoreGetProto, StoreNew, StoreSet, StoreSetProto};
+use substreams_solana::pb::sol as solana;
+
+#[substreams::handlers::map]
```
robinbernon, line 10: Perhaps you could move this to a modules folder like the other substreams are now doing?

spl-balance/substreams.yaml
```diff
@@ -0,0 +1,36 @@
...
+    - sol_token.proto
+  importPaths:
+    - ../common/proto
+
+binaries:
+  default:
+    type: wasm/rust-v1
+    file: "../target/wasm32-unknown-unknown/release/spl_balance.wasm"
+
+modules:
```
robinbernon, line 20: Think these modules are back to front?

spl-balance/src/lib.rs
```diff
@@ -0,0 +1,178 @@
...
+                            Ok(instruction::TokenInstruction::InitializeMint {
+                                decimals,
+                                mint_authority,
+                                freeze_authority,
+                            }) => {
+                                log::info!("Instruction: InitializeMint");
+                                output.set(
+                                    0,
+                                    format!(
+                                        "address:{}",
```
robinbernon, line 94: Maybe use a storekey struct or a small fn that you can reuse between modules

spl-balance/src/option.rs
```diff
@@ -0,0 +1,951 @@
...
+
+use std::{
+    convert, mem,
+    ops::{Deref, DerefMut},
+};
+
+/// A C representation of Rust's `std::option::Option`
+#[repr(C)]
+#[derive(Copy, PartialEq, PartialOrd, Eq, Ord, Debug, Hash)]
+pub enum COption<T> {
```
robinbernon, line 15: This whole file can be removed and you can switch back to using std::Option instead

spl-balance/src/instruction.rs
robinbernon, line 8: Shrink this enum down to having just three fields: 

 - InitializeMint
 - InitializeMint2
 - NonMintInstruction

You will also be able to remove all C api code and a bunch of the map code

spl-balance/src/instruction.rs
```diff
@@ -0,0 +1,658 @@
...
+            }
+            _ => Err(Error::Unexpected(format!("Invalid Instruction"))),
+        }
+    }
+}
+
+/// Specifies the authority type for SetAuthority instructions
+#[repr(u8)]
+#[derive(Clone, Debug, PartialEq)]
+pub enum AuthorityType {
```
robinbernon, line 637: You can remove this also

spl-balance/src/lib.rs
```diff
@@ -0,0 +1,178 @@
...
+    freeze_authority_opt: Option<Vec<u8>>,
+    tx_id: String,
+    block_height: u64,
+) -> proto::TokenAccount {
+    let mut token = proto::TokenAccount {
+        address: bs58::encode(&token_address).into_string(),
+        name: "".to_string(),
+        symbol: "".to_string(),
+        decimals: decimal.into(),
+        freeze_authority: "".to_string(),
```
robinbernon, line 154: You could have this all on one line using .map() and then .unwrap_or_default(). This would then make it simple enough to doing the instantiation back inside the main module code rather than this fn here.

spl-balance/src/lib.rs
```diff
@@ -0,0 +1,178 @@
...
+    token_address: Vec<u8>,
+    decimal: u8,
+    mint_authority: Vec<u8>,
+    freeze_authority_opt: Option<Vec<u8>>,
+    tx_id: String,
+    block_height: u64,
+) -> proto::TokenAccount {
+    let mut token = proto::TokenAccount {
+        address: bs58::encode(&token_address).into_string(),
+        name: "".to_string(),
```
robinbernon, line 151: Think all protos inherit the default trait automatically so you might be able to get away with ..default at the bottom rather than manually spefying these fields during the instantiation

spl-balance/src/lib.rs
```diff
@@ -0,0 +1,178 @@
...
+    };
+
+    // set freeze authority if it exists
+    if freeze_authority_opt.is_some() {
+        token.freeze_authority = bs58::encode(&freeze_authority_opt.unwrap()).into_string();
+    }
+    return token;
+}
+
+// pull the TokenAccount using the key provided
```
robinbernon, line 167: What are your plans for this code block?

spl-balance/src/pb.rs
```diff
@@ -0,0 +1,9 @@
+#[rustfmt::skip]
+#[path = "../target/pb/messari.sol_token.v1.rs"]
+pub(in crate::pb) mod sol_token_v1;
```
robinbernon, line 3: Just noting this here for the future but what we should do here is filter out all protos which are from substream-helpers in the build.rs script and not build bindings for them. Then we can just import the bindings from the substreams-helper crate instead which will avoid the unnecessary duplication of code here

spl-balance/Cargo.toml
```diff
@@ -0,0 +1,23 @@
+[package]
+name = "spl-balance"
+version = "0.1.0"
+description = "Messari's standardized substream for solana SPL tokens"
+edition = "2021"
+repository = "https://github.com/messari/substreams"
```
robinbernon, line 6: Not really that important but it would be nice to have all these links be consistant. (If you use the CLI to create the substream projects it will do this for you :) )

PR #129, opened 2023-01-31
chore(); docs; add walkthrough.md

dmelotik:
Added this file to describe some of our terminology for others to reference


yulesa (commented):
I like it. 

Maybe I would highlight a little better which ones are the most important files, and what they contain. It's in the text but I would repeat on structure to make it clear or easier when browsing with eyes.

docs/WALKTHROUGH.md
```diff
@@ -0,0 +1,21 @@
...
+
+Our project is made up of substream modules. Each directory (for example `./eth-balance`) contains a group of modules (1 or more). That group of modules describes data around a central theme. If it contains a `README.md` you can learn more there, or by understanding the manifest file, `substreams.yaml` (described [here](https://substreams.streamingfast.io/developers-guide/creating-your-manifest)).
+
+### `substreams.yaml`
+
+This is the manifest file. it describes the substream modules that make up the substream.
+
+### `./src/*.rs`
+
+This is where the mappings are stored. These are the functions that take the input and transform it into the output defined in the manifest file.
```
robinbernon, line 15: Maybe refer to modules folder rather than src folder here

PR #131, opened 2023-02-23
eth-supply: Added substream to calculate total ETH supply

jaimehgb:
ETH supply can go up and down in the following ways:
- Genesis block: The first block of the chain minted ~72M ETH, from its crowdsale. We are calculating these from all balance changes inside the only transaction in that block (0).
- Miner rewards: When minting a new block, miners get freshly minted ETH. This comes in a Block.BalanceChange in the pb Block. 
- Uncle rewards: As an incentive for miners not to build empty blocks, and because of the probability of creating orphan blocks, miners get ETH rewards for uncle blocks included in the block headers. These are on top of the default miner rewards.
- EIP1559 burn: After EIP1559, transaction fees are divided in base fee + priority fee. The base fee is burned. 

After the merge, ETH minted by the beacon chain also adds to the total supply, but this data is currently not available, so we are not accounting for it yet.


eth-supply/proto/v1/eth_supply.proto
```diff
@@ -0,0 +1,11 @@
+syntax = "proto3";
+
+package messari.eth_supply.v1;
+
+message EthSupply {
+  string genesis = 1;      // string: BigInt, in wei
```
robinbernon, line 6: Is there a reason why you're not using a BigInt type here and are using string instead?

eth-supply/src/helpers.rs
```diff
@@ -0,0 +1,48 @@
...
+
+// Aux function to calculate the actual change in balance.
+pub fn get_delta(change: pbeth::v2::BalanceChange) -> BigInt {
+    let old = pb_bigint_to_bigint(change.old_value);
+    let new = pb_bigint_to_bigint(change.new_value);
+    new.sub(old)
+}
+
+// Aux function to convert a protobuf BigInt into a useable BigInt.
+fn pb_bigint_to_bigint(bi: Option<pbBigInt>) -> BigInt {
```
robinbernon, line 44: This def works fine as is, although I'm just thinking, as I've already implemented the following: 
https://github.com/messari/substreams/pull/109/files#diff-8d985f93a4120bd236f857ccc3b97c14c19c9d81e56d3a18f2830ac5476ca129
In file: network/src/utils.rs
Should we try to standardise and code these conversions ih the same way?

eth-supply/src/helpers.rs
```diff
@@ -0,0 +1,48 @@
...
+// ETH burns the base fee. So burn per block = base fee = baseFeePerGas * gasUsed
+pub fn get_eip1559_burn(header: &BlockHeader) -> BigInt {
+    return pb_bigint_to_bigint(header.base_fee_per_gas.to_owned())
+        .mul(BigInt::from(header.gas_used));
+}
+
+// If passed the genesis block it will calculate all value minted from the crowdsale
+// on that block (it doesn't come as a block BalanceChange, so we inspect the calls inside)
+// the only transaction in that block.
+pub fn get_genesis_value(block: &pbeth::v2::Block) -> BigInt {
```
robinbernon, line 20: Quite similar to https://github.com/messari/substreams/pull/109/files#diff-4ad539ed90583be7660cef0adf5e3fb5f8956c713edcab22c06dc06b526ea727
network/src/block_handler.rs
supply and rewards fns 
Think it might be a good idea to add general block logic like this to a shared block_hander  tool like the one I built already in a common lib and use it like that instead?

eth-supply/src/lib.rs
```diff
@@ -0,0 +1,103 @@
...
+#[substreams::handlers::map]
+fn map_supply_delta(
+    block: pbeth::v2::Block,
+) -> Result<eth_supply::EthSupply, substreams::errors::Error> {
+    let mut block_rewards: BigInt = BigInt::from(0);
+    let mut uncle_rewards: BigInt = BigInt::from(0);
+    let genesis: BigInt = get_genesis_value(&block);
+
+    for change in block.balance_changes {
+        match pbeth::v2::balance_change::Reason::from_i32(change.reason).unwrap_or_default() {
```
robinbernon, line 35: As mentioned above about, have a look at my block_hander in the network PR

eth-supply/src/lib.rs
```diff
@@ -0,0 +1,103 @@
...
+                uncle_rewards = uncle_rewards.add(get_delta(change));
+            }
+            pbeth::v2::balance_change::Reason::RewardMineBlock => {
+                block_rewards = block_rewards.add(get_delta(change));
+            }
+            _ => {}
+        }
+    }
+
+    let burned = get_eip1559_burn(&block.header.unwrap());
```
robinbernon, line 46: For stuff like this use block.header.as_ref().unwrap() instead

eth-supply/src/lib.rs
```diff
@@ -0,0 +1,103 @@
+#[rustfmt::skip]
```
robinbernon, line 1: For the modules could you use a modules folder layout instead? Have a look at some of the other substreams in our repo to see what I mean here

eth-supply/src/lib.rs
```diff
@@ -0,0 +1,103 @@
...
+    Ok(deltas)
+}
+
+const GENESIS_BALANCE_STORE_KEY: &str = "genesis";
+const MINT_BALANCE_STORE_KEY: &str = "mint";
+const UNCLE_BALANCE_STORE_KEY: &str = "uncles";
+const BURNED_BALANCE_STORE_KEY: &str = "burned";
+
+#[substreams::handlers::store]
+fn store_supply(deltas: eth_supply::EthSupply, output: store::StoreAddBigInt) {
```
robinbernon, line 63: You don't need to store all the values here. Instead you only need to be tracking the total. You can get all the other values by making map_output consume from the output of map_supply_delta

eth-supply/src/lib.rs
```diff
@@ -0,0 +1,103 @@
...
+    output.add(
+        0,
+        BURNED_BALANCE_STORE_KEY,
+        BigInt::from_str(deltas.burned.as_str()).unwrap(),
+    );
+}
+
+#[substreams::handlers::map]
+fn map_output(
+    store_supply: store::StoreGetBigInt,
```
robinbernon, line 88: As mentioned above - use the output from map_supply_delta as an input here and then supplement the struct with the total field value that you have calculated in the store

robinbernon (changes requested):

eth-supply/src/helpers.rs
jaimehgb: Yes! It is an operation we are going to need to do all the time. I can move it to substreams_helper. Will use your implementation instead.

eth-supply/src/lib.rs
jaimehgb: Nothing against it, but curious about the why :P

eth-supply/proto/v1/eth_supply.proto
robinbernon: Yeh you can just define your own as a bytes object (It's been done like that for some of the other substreams for if you want to see an example).

eth-supply/src/lib.rs
robinbernon: So what you had before is fine and would work however it's best to use this way as a good habbit. The reason is that if you don't use .as_ref() before you unwrap an option you take ownership over the option. So your way essentially takes ownership of the field and then takes a reference again from the moved value. Under the hood I'm sure the compiler would optimise out the difference in implementation however in other scenarios this could give you various errors with the borrow checker when it comes to the taking of ownership here. So yes this works although i'd say as_ref() would be better practise here

eth-supply/substreams.yaml
```diff
@@ -0,0 +1,42 @@
...
+  - name: store_supply
+    kind: store
+    updatePolicy: add
+    valueType: bigint
+    inputs:
+      - map: map_supply_delta
+
+  - name: map_output
+    kind: map
+    initialBlock: 0
```
robinbernon, line 32: 0 is the default here. Only want to specify when wanting non zero starting blocks

eth-supply/src/helpers.rs
jaimehgb: Oh! This is useful. I'll switch my logic to get new supply, uncles, and burnt fees to use that instead.

The genesis issuance, do you think it is worth it? It is a one off thing, only happened on the very first block. I think it is too specific of a usecase to generalize that way. If someone needs to calculate genesis issuance again they can just import `get_genesis_value` instead?

eth-supply/src/lib.rs
jaimehgb: Done!

eth-supply/src/helpers.rs
jaimehgb: Copied the trait you have there to convert between `scalar::BigInt <> eth::BigInt`. Then the ones to convert between `scalar::BigInt <> crate::pb::BigInt` I left on my own crate `helpers.rs`

eth-supply/src/abi.rs
```diff
@@ -0,0 +1 @@
+// DO NOT EDIT - the file is generated by build script
```
robinbernon, line 1: If no ABI attached can you just get rid of this file?

eth-supply/build.rs
```diff
@@ -0,0 +1,10 @@
+use anyhow::{Ok, Result};
+use substreams_common::codegen;
+
+fn main() -> Result<(), anyhow::Error> {
+    println!("cargo:rerun-if-changed=proto");
+    println!("cargo:rerun-if-changed=abi");
```
robinbernon, line 6: No abi for this project right? Can you drop this line?

robinbernon (approved):

PR #132, opened 2023-02-27
Parquet Sink for substreams

robinbernon:


parquet/src/file.rs
```diff
@@ -0,0 +1,64 @@
...
+    pub(crate) fn new(file_data: Vec<u8>, output_location: Location) -> File {
+        File {
+            file_data,
+            output_location,
+        }
+    }
+}
+
+pub(crate) enum Location {
+    DataWarehouse(PathBuf),
```
jaimehgb, line 18: I might be getting ahead since the PR is still a draft, or I might just be too noob with rust to properly understand what's going on😅 If so, call me out.

But should we make a DataWarehouse specific location? I feel like the sink should not know about the different destination types. Can we generalize it in some destination agnostic way?

If `process_substream` depends on a `trait FileSaver { save_file(&self, file: File) }` for example, that we can implement separately and handles the different storage types that a user might need. Implementations for `FileSaver` would live outside of the sink crate.

robinbernon: So the sink ingests data from SF and is responsible for producing parquet files as an output. The only thing you're ever gonna need to do with these files is save them somewhere. So yes we could have an intermediatory stream that just spits out Vec<u8> to represent the file data and another service that then wraps them in this file struct, however I feel there is currently no need to do this as we don't need this intermediatory stream for any other purpose right now.

parquet/src/decoder.rs
```diff
@@ -0,0 +1,245 @@
...
+                field_decoder.decode(data, wire_type, uncompressed_file_size, current_definition_lvl, last_repetition_lvl)
+            }
+            Decoder::StructDecoder(struct_decoder) => struct_decoder.decode(data, wire_type, uncompressed_file_size, current_definition_lvl, last_repetition_lvl),
+        }
+    }
+
+    pub(crate) fn new(proto_descriptors: &Vec<FileDescriptorProto>, field_descriptor: &FieldDescriptorProto, parquet_schema_builder: &mut ParquetSchemaBuilder, track_definition_lvls: bool, track_repetition_lvls: bool) -> Decoder {
+        let field_type = field_descriptor.r#type.unwrap() as Type;
+
+        match field_type {
```
dmelotik, line 42: Why can't we just use a generic type here?

messari-cli/src/commands/upload_cli_to_aws.rs
```diff
@@ -0,0 +1,35 @@
...
+
+        let file_data = fs::read(&cli_file_path).unwrap();
+        let file_name = cli_file_path.file_name().unwrap().to_string_lossy().to_string();
+        assert_eq!(file_name.as_str(), "messari", "CLI needs to be called messari!");
+
+        let bucket_name = "spkg-bucket";
+        let region = "us-west-2".parse().unwrap();
+        let credentials = Credentials::default().unwrap();
+        let bucket = Bucket::new(bucket_name, region, credentials).unwrap();
+        let response_data = bucket.put_object("/messari-cli/messari", file_data.as_slice()).await.unwrap();
```
jaimehgb, line 31: Same as the path is a config parameter, shouldn't the bucket name, region, and full path be config too?

Same for the spkgs upload.

robinbernon: I can put in an option cmd arg for non messari people if they want to override these settings although I like the idea that it's super simple for messari people to use the CLI (they don't have to fish through docs to find the right bucket info etc)

messari-cli/src/streaming_fast/process_substream.rs
```diff
@@ -16,12 +18,50 @@ use crate::streaming_fast::streamingfast_dtos::ForkStep::StepIrreversible;
...
+        chain_override
+    } else {
+        get_chain_info(&package)
+    };
+
+    let (mut sink, proto_type_name) = get_sink_and_proto_type_name(&package, config.output_module.as_str(), encoding_type, location_type, data_location_path, &chain);
+
+    if let Some(substream_name) = config.substream_name_override {
+        package.package_meta.iter_mut().next().unwrap().name = substream_name;
+    }
```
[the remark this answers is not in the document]
robinbernon: That's true if only the name is changed.. however a fork will always get different data and be processed differently to another fork due to diff accompanying param overrides so I don't think there is need to worry about this?

messari-cli/src/streaming_fast/process_substream.rs
```diff
@@ -0,0 +1,303 @@
...
+            if block_num > latest_block_num {
+                latest_block_num = block_num;
+            }
+        }
+        return latest_block_num;
+    }
+
+    // Otherwise we will fall back to taking the initial block number specified for the given module
+    for module in package.modules.as_ref().unwrap().modules.iter() {
+        if module.output.as_ref().unwrap().r#type.as_str() == proto_type_name {
```
jaimehgb, line 249: This unwrap seems to fail when there is a store module in the manifest (since stores don't have output)

PR #133, opened 2023-03-15
Added subgraph for eth-supply substream

jaimehgb:


eth-supply/src/modules/4_map_entity_changes.rs
```diff
@@ -0,0 +1,72 @@
...
+
+use crate::pb::eth_supply::v1::EthSupply;
+
+#[substreams::handlers::map]
+fn map_entity_changes(
+    block_delta: EthSupply,
+    cumulative: EthSupply,
+) -> Result<EntityChanges, substreams::errors::Error> {
+    let hash_bytes: &Vec<u8> = cumulative.block_hash.as_ref();
+    let block_hash: String = hash_bytes.encode_hex::<String>();
```
robinbernon, line 18: If you're converting the hash to a string each time you might as well do that in module 1 and pass it in the proto in string form rather than in bytes

jaimehgb: I need both the string and bytes format for the EntityChange. If I passed it as a string, I'd have to convert it to bytes too 😅 

The ID field in the subgraph schema will be the block hash, and EntityChange requires the ID to be a string. But the block hash field is of type bytes. 

I've made some changes though, since we can get the block number and hash from `sf.substreams.v1.Clock` we don't need to pass it explicitly in the proto.

robinbernon (dismissed):

robinbernon (approved):

PR #135, opened 2023-04-12
Added EventHandler and Storage utils to substreams-helper

jaimehgb:
> **NOTE:** This PR is opened against the `eth-supply` branch because of some dependencies from it. I think it is easier to review this changes in isolation than if everything was mixed.

### Summary 

This PR adds a few utilities for dealing with events and storage changes.

### Events
In subgraphs, handling an event is as simple as writing a function that takes the event as an argument. Here in substreams, so far what we've been doing is to reimplement over and over all the iterating over the logs in the block and matching their signature with our ABI event.

I've added a very simple class (can you call them classes in rust?) to handle all that. The usage is documented in the code, but it is something like:
```
let mut some_handler = |ev: SomeEvent, tx: &TransactionTrace, log: &Log| { /*...*/ };

let eh = EventHandler::new(&block);
eh.on::<SomeEvent, _>(some_handler);
eh.handle_events();
``` 

### Storage Changes
On top of that, I've had to work with `storage_changes`. Working with them is also very raw at the moment, it involves abi_encoding types, hashing stuff and things like that. I started a very naive set of utilities to work with that. Its purpose, more than being usable, is to have us add to that instead of implementing our own logic on each individual substream every time.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -155,7 +155,7 @@ dependencies = [
  "anyhow",
  "ethabi",
  "num-bigint",
- "prost 0.11.6",
+ "prost 0.11.9",
  "substreams 0.5.6",
  "substreams-common",
  "substreams-ethereum",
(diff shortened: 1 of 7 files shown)


jaimehgb:
Shamelessly merging without approval, to unblock other dependants. Can review "a posteriori".

PR #139, opened 2023-05-22
Added storage utils for arrays and structs. Supports recursivity (arrays of structs of ints for example)

jaimehgb:


substreams-helper/src/common.rs
```diff
@@ -61,3 +61,9 @@ impl HasAddresser for StoreGetInt64 {
         self.get_last(key.to_hex()).is_some()
     }
 }
+
+impl HasAddresser for Address {
```
robinbernon, line 65: Is this just impled for Addresses only? If so perhaps you could change the store args below to have name address rather than store and type as Address rather than impl Store to make it more clear what is compared with what here?

substreams-helper/src/storage.rs
```diff
@@ -1,31 +1,121 @@
...
             if !store.has_address(Address::from_slice(change.address.as_slice())) {
                 continue;
             }
-            changes.push(change.to_owned());
+
+            let mut preimage = None;
+            let preimage_bytes = call
+                .call
+                .keccak_preimages
+                .get(&substreams::Hex::encode(change.key.clone()));
```
robinbernon, line 94: Forgot the specifics but could you use Hex::__ with some reference passing instead of a clone here?

substreams-helper/src/storage.rs
```diff
@@ -1,31 +1,121 @@
...
+            if preimage_bytes.is_some() {
```
robinbernon, line 95: Use: 
if let Some(preimage) = preimage_bytes {
    ....
}
Instead for this

substreams-helper/src/storage.rs
```diff
@@ -1,31 +1,121 @@
...
 pub fn get_storage_changes_for_addresses(
     store: &impl HasAddresser,
     block: &pbeth::v2::Block,
-) -> Vec<pbeth::v2::StorageChange> {
-    let mut changes: Vec<pbeth::v2::StorageChange> = vec![];
+) -> Vec<StorageChange> {
+    let mut changes: Vec<StorageChange> = vec![];
     for call in block.calls() {
         for change in &call.call.storage_changes {
             if !store.has_address(Address::from_slice(change.address.as_slice())) {
```
robinbernon, line 86: Doing the if statement the other way round will save you a negation operation and a continue line :)

substreams-helper/src/storage.rs
```diff
@@ -1,31 +1,121 @@
...
+            changes.push(StorageChange {
+                change: change.to_owned(),
+                preimage: preimage,
+            });
         }
     }
     changes
 }
 
+pub fn keccak256(data: Vec<u8>) -> Vec<u8> {
```
robinbernon, line 107: Just thinking - would it be possible to set the signature as [0u8; 32] instead of Vec<u8> as it's more informative of the kekak data? Your abi_decode fn can take T:Into<Vec<u8>> instead or something similar to keep the logic the same if that works for you?

substreams-helper/src/storage.rs
```diff
@@ -34,12 +124,317 @@ impl Uint256 {
     }
 }
 
+impl Default for Uint256 {
```
robinbernon, line 127: Does BigInt have a default impled on it already? If so you can just derive Default for the struct rather than manually impling it

substreams-helper/src/storage.rs
robinbernon: Just checked and it looks like you could even just accept slices instead based on the current impls of the trait?

substreams-helper/src/storage.rs
```diff
@@ -34,12 +124,317 @@ impl Uint256 {
...
+    pub offset: usize,
+}
+
+impl Uint128 {
+    pub fn storage_key(&self) -> Vec<u8> {
+        ethabi::encode(&[self.slot.abi_token()])
+    }
+}
+
+impl Default for Uint128 {
```
robinbernon, line 182: Same here as above :)

substreams-helper/src/storage.rs
```diff
@@ -34,12 +124,317 @@ impl Uint256 {
...
+            )));
+        }
+
+        if let Some(offset_override) = offset {
+            self.offset = offset_override;
+        }
+
+        let raw = subslot(slots[0].clone(), self.size(), self.offset);
+        let decoded = BigInt::abi_decode(raw);
+        if decoded.is_err() {
```
robinbernon, line 215: Better semantics might be to re-write using a match here

substreams-helper/src/storage.rs
```diff
@@ -34,12 +124,317 @@ impl Uint256 {
...
+    /// Will return a list of all storage changes involving this array, from a given list of changes.
+    pub fn filter_array_changes(
+        &self,
+        changes: Vec<StorageChange>,
+        expected_array_len: BigInt,
+    ) -> Vec<StorageChange> {
+        let mut filtered_changes = vec![];
+        for change in changes {
+            let index = self.infer_array_index_from_storage_key(change.change.key.clone());
+            if index.is_none() {
```
robinbernon, line 350: if let Some(_) instead here :)

substreams-helper/src/storage.rs
```diff
@@ -34,12 +124,317 @@ impl Uint256 {
...
+        if slot_diff.modulo(slots_per_item.as_ref()) != BigInt::zero() {
+            return None;
+        }
+
+        Some(slot_diff / slots_per_item)
+    }
+
+    /// Returns the EVM storage key where the value of a given array index is stored.
+    pub fn storage_key_at_index(&self, index: BigInt) -> Vec<u8> {
+        let slots_per_item = BigInt::from(size_to_slots(self.item.size()));
```
robinbernon, line 381: size_to_slots here is usize. I think it would be fine to write the modulo logic to accept a usize arg instead and return a usize arg. Will save on the BigInt wrapping :)

substreams-helper/src/storage.rs
```diff
@@ -34,12 +124,317 @@ impl Uint256 {
...
+
+pub struct StructField {
+    pub name: String,
+    value: Box<dyn StorageLayout>,
+    offset: usize,
+    start_slot: usize,
+    end_slot: usize,
+}
+
+pub struct Struct {
```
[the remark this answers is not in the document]
robinbernon: Realised maybe something like EvmStruct may be more appropriate than DynamicStruct

substreams-helper/src/common.rs
jaimehgb: It is also implemented for all substream stores. So if you have a Store with addresses of contracts it works too.

robinbernon: Didn't realise. Ok makes sense, thanks 👍

substreams-helper/src/storage.rs
jaimehgb: I'll still need to wrap the usize in a BigInt, since I need to divide a BigInt by the usize in the calculation of the modulo. It will just be done lower down the stack.

Changed it, the implementation of the modulo now looks something like:
```
impl<T> NumberModulo<T> for BigInt
where
    T: Into<BigInt> + Clone,
{
    fn modulo(&self, other: T) -> Self {
        let division = self.to_owned() / other.clone();
        self.to_owned() - division * other
    }
}
```

Types implementing `Into<BigInt>` are already usable in divisions and multiplications.

substreams-helper/src/storage.rs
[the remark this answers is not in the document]
jaimehgb: EvmStruct works for me. Agree that Struct alone can be confusing :P 

Changed!

substreams-helper/src/storage.rs
jaimehgb: Doesn't seem like it:
```
the trait bound `substreams::scalar::BigInt: Default` is not satisfied
the trait `Default` is not implemented for `substreams::scalar::BigInt
```

Will keep it in mind for the next times though.

substreams-helper/src/storage.rs
jaimehgb: Yup, just found it easier to work with vecs 😅 

Changed to slices

substreams-helper/src/storage.rs
jaimehgb: Same as above, prefer to return early / continue than keep nesting things 😬 

Changed though, if it is more idiomatic for rust.

substreams-helper/src/storage.rs
jaimehgb: Used to early returns. Changed :P

substreams-helper/src/storage.rs
jaimehgb: I can make it work with something like:
```
.get(&hex::encode(change.key.as_slice()));
```

